PR #61, opened 2015-02-13
Added new rule for newline at end of file

### LINKED ISSUE #54: Add final newline to code files
More of a git formatting rule but mentioned in http://channel9.msdn.com/Blogs/dotnet/github-for-the-roslyn-team

khellang:
This does the following:
  1. Removes all trailing EndOfLine trivia from the last node of the document.
  2. Replaces all leading EndOfLine trivia on the EndOfFile token with a single CarriageReturnLineFeed trivia.

Fixes #54

### DIFF
--- a/src/Microsoft.DotNet.CodeFormatting.Tests/Rules/HasNewLineAtEndOfFileFormattingRuleTests.cs
+++ b/src/Microsoft.DotNet.CodeFormatting.Tests/Rules/HasNewLineAtEndOfFileFormattingRuleTests.cs
@@ -1,0 +1,65 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
+
+namespace Microsoft.DotNet.CodeFormatting.Tests
+{
+    public class HasNewLineAtEndOfFileFormattingRuleTests : SyntaxRuleTestBase
+    {
+        internal override ISyntaxFormattingRule Rule
+        {
+            get { return new Rules.HasNewLineAtEndOfFileFormattingRule(); }
+        }
+
+        [Fact]
+        public void ShouldAddNewLine()
+        {
+            var source = "public class C { }";
+
+            var expected = "public class C { }\r\n";
+
+            Verify(source, expected);
+        }
+
+        [Fact]
+        public void ShouldRemoveExtraNewLines()
+        {
+            var source = "public class C { }\r\n\r\n\n\r\n\n\r";
+
+            var expected = "public class C { }\r\n";
+
+            Verify(source, expected);
+        }
+
+        [Fact]
+        public void ShouldNotCareAboutExistingCarriageReturn()
+        {
+            var source = "public class C { }\r";
+
+            var expected = "public class C { }\r\n";
+
+            Verify(source, expected);
+        }
+
+        [Fact]
+        public void ShouldNotCareAboutExistingLineFeed()
+        {
+            var source = "public class C { }\n";
+
+            var expected = "public class C { }\r\n";
+
+            Verify(source, expected);
+        }
+
+        [Fact]
+        public void ShouldHandleEmptyDocument()
+        {
+            var source = string.Empty;
+
+            var expected = "\r\n";
+
+            Verify(source, expected);
+        }
+    }
+}
--- a/src/Microsoft.DotNet.CodeFormatting/Rules/HasNewLineAtEndOfFileFormattingRule.cs
+++ b/src/Microsoft.DotNet.CodeFormatting/Rules/HasNewLineAtEndOfFileFormattingRule.cs
@@ -1,0 +1,51 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+
+namespace Microsoft.DotNet.CodeFormatting.Rules
+{
+    [SyntaxRuleOrder(SyntaxRuleOrder.HasNewLineAtEndOfFile)]
+    internal sealed class HasNewLineAtEndOfFileFormattingRule : ISyntaxFormattingRule
+    {
+        public SyntaxNode Process(SyntaxNode syntaxRoot)
+        {
+            var node = syntaxRoot.DescendantNodes().LastOrDefault();
+
+            if (node != null)
+            {
+                syntaxRoot = syntaxRoot.ReplaceNode(node, RemoveNewLines(node));
+            }
+
+            var token = syntaxRoot.DescendantTokens().Single(x => x.IsKind(SyntaxKind.EndOfFileToken));
+
+            return syntaxRoot.ReplaceToken(token, AdjustNewLines(token));
+        }
+
+        private static SyntaxNode RemoveNewLines(SyntaxNode node)
+        {
+            var newTrivia = Enumerable.Empty<SyntaxTrivia>();
+
+            if (node.HasTrailingTrivia)
+            {
+                newTrivia = node.GetTrailingTrivia().Where(x => !x.IsKind(SyntaxKind.EndOfLineTrivia));
+            }
+
+            return node.WithTrailingTrivia(newTrivia);
+        }
+
+        private static SyntaxToken AdjustNewLines(SyntaxToken token)
+        {
+            var newTrivia = Enumerable.Empty<SyntaxTrivia>();
+
+            if (token.HasLeadingTrivia)
+            {
+                newTrivia = token.LeadingTrivia.Where(x => !x.IsKind(SyntaxKind.EndOfLineTrivia));
+            }
+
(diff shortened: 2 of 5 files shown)


src/Microsoft.DotNet.CodeFormatting.Tests/Rules/HasNewLineAtEndOfFileFormattingRuleTests.cs
```diff
@@ -0,0 +1,65 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Xunit;
+
+namespace Microsoft.DotNet.CodeFormatting.Tests
+{
+    public class HasNewLineAtEndOfFileFormattingRuleTests : SyntaxRuleTestBase
```
jaredpar: `internal sealed`

src/Microsoft.DotNet.CodeFormatting/Rules/HasNewLineAtEndOfFileFormattingRule.cs
```diff
@@ -0,0 +1,51 @@
...
+        public SyntaxNode Process(SyntaxNode syntaxRoot)
+        {
+            var node = syntaxRoot.DescendantNodes().LastOrDefault();
+
+            if (node != null)
+            {
+                syntaxRoot = syntaxRoot.ReplaceNode(node, RemoveNewLines(node));
+            }
+
+            var token = syntaxRoot.DescendantTokens().Single(x => x.IsKind(SyntaxKind.EndOfFileToken));
```
jaredpar: CC @jasonmalinowski because he will have better knowledge of the parse tree here.  My concern is whether or not we are guaranteed the `EndOfFileToken` will be the only descendant token here in all cases.  If not `Single` will throw and break the formatter.

src/Microsoft.DotNet.CodeFormatting/Rules/HasNewLineAtEndOfFileFormattingRule.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            var newTrivia = Enumerable.Empty<SyntaxTrivia>();
+
+            if (token.HasLeadingTrivia)
+            {
+                newTrivia = token.LeadingTrivia.Where(x => !x.IsKind(SyntaxKind.EndOfLineTrivia));
+            }
+
+            return token.WithLeadingTrivia(newTrivia.AddNewLine());
+        }
```
jaredpar: The actual formatting step, invoking the Formatter, will remove the final new line in a file.  This is an uncontrollable option of the formatter at this time.  Hence adding this token here will just result in it being removed later.  I think we should just return token directly.

src/Microsoft.DotNet.CodeFormatting/Rules/HasNewLineAtEndOfFileFormattingRule.cs
```diff
@@ -0,0 +1,51 @@
...
+            return node.WithTrailingTrivia(newTrivia);
+        }
+
+        private static SyntaxToken AdjustNewLines(SyntaxToken token)
+        {
+            var newTrivia = Enumerable.Empty<SyntaxTrivia>();
+
+            if (token.HasLeadingTrivia)
+            {
+                newTrivia = token.LeadingTrivia.Where(x => !x.IsKind(SyntaxKind.EndOfLineTrivia));
```
jaredpar: This needs to be a bit more nuanced here.  Consider the following code:

``` csharp
class C
{

}
// comment 1
// comment 2 
```

The result of running this rule would be 

``` csharp
class C
{

}
// comment 1 // comment 2
``

This is because the leading trivia can include comments (both structured and unstructured).  

I had to make a similar change for braces above a new line.  It has virtually the same problems as this particular change.  May want to just make this a helper method. 

https://github.com/dotnet/codeformatter/blob/master/src/Microsoft.DotNet.CodeFormatting/Rules/BraceNewLineRule.cs#L142

src/Microsoft.DotNet.CodeFormatting/Rules/HasNewLineAtEndOfFileFormattingRule.cs
jasonmalinowski: @basoundr @heejaechang would know more here than I.

khellang: @jaredpar It's not the only descendant token, it's getting _a single_ EndOfFile token. Wouldn't it be weird if there were multiple EOFs?

jaredpar: @khellang ah you are correct.  I wasn't reading the conditional on `Single` correctly.

khellang: @jaredpar I could of course change it to `Last`, just in case?

src/Microsoft.DotNet.CodeFormatting/Rules/HasNewLineAtEndOfFileFormattingRule.cs
khellang: Yeah, I'll get more tests in there to try to nail more edge cases :smile:

src/Microsoft.DotNet.CodeFormatting/Rules/HasNewLineAtEndOfFileFormattingRule.cs
khellang: Not sure what you mean by "adding the token". Do you mean the CRLF trivia? Doesn't the tests invoke the formatter?

src/Microsoft.DotNet.CodeFormatting/Rules/HasNewLineAtEndOfFileFormattingRule.cs
jaredpar: @khellang I actually like it much better as written.  If there is more than one EndOfFileToken then that is a bug.  Want to make sure we seek it ;)

jaredpar: So yes I did start this comment thread asking for a change to the code and end with asking for it to not change.

src/Microsoft.DotNet.CodeFormatting/Rules/HasNewLineAtEndOfFileFormattingRule.cs
jaredpar: @khellang just remember the golden rule of editing trivia in Roslyn

> Trivia is never trivial 

It's the Roslyn motto ;)

src/Microsoft.DotNet.CodeFormatting/Rules/HasNewLineAtEndOfFileFormattingRule.cs
jaredpar: Correct, I meant adding the newline.  I was asking why we add it here knowing the formatter is just going to remove it soon after.

src/Microsoft.DotNet.CodeFormatting/Rules/HasNewLineAtEndOfFileFormattingRule.cs
khellang: Great :+1:

src/Microsoft.DotNet.CodeFormatting/Rules/HasNewLineAtEndOfFileFormattingRule.cs
khellang: So... That's why I asked if the tests run the formatter. I think I saw it in there somewhere :smile: They get the correct newline at the end. I also ran this over the Nancy and scriptcs codebase, and it seemed to work nicely.

src/Microsoft.DotNet.CodeFormatting/Rules/HasNewLineAtEndOfFileFormattingRule.cs
khellang: LOL. Good to hear it's not just me that find this thing terrifying :wink:

src/Microsoft.DotNet.CodeFormatting/Rules/HasNewLineAtEndOfFileFormattingRule.cs
jaredpar: @khellang correct it will be removed.  But leaving it in this code can give the false impression to the casual reader that the intent is to have all files in end in a newline.  

I know because I fell into this trap with a previous version of the code base ;)

khellang: _Isn't_ the intent that we want all files to end in a newline? And if it's removed, how come it's present in all flies of a sln after running it through the CodeFormatter with this rule enabled?

PR #62, opened 2015-02-13
Add new rule to convert Test Assert.True or False with redundant conditi...

YingP99:
Add new rule to convert Test Assert.True or False with redundant condition.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -28,4 +28,4 @@ IF EXIST "%~dp0src\packages" goto build
 
 :build
 
-%BUILD_TOOLS_PATH% %SOLUTION_PATH% /p:OutDir="%~dp0bin " /nologo /m /v:m /flp:verbosity=normal %*
+%BUILD_TOOLS_PATH% %SOLUTION_PATH% /p:OutDir="%~dp0bin" /nologo /m /v:m /flp:verbosity=normal %*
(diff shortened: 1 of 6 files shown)


ellismg:
Can you redo this commit as yourself?  Set core.name and core.e-mail to your information and then git commit --amend --reset-author.  Then force push your branch with git push origin +assertrule:assertrule.

YingP99:
@ellismg Done!

jaredpar:
I'm not sure this rule is appropriate for the code formatter for a couple of reasons:

- It is not a part of our coding guidelines.  All rules today which aren't a part of the guidelines are being moved out and into separate projects.
- This is not a rule that can be run safely on a code base.  At least not without introducing semantic analysis into the equation.  The translation between `Assert.True(x ==y)` to `Assert.Equals(x, y)` is not always correct.  It confuses reference and value equality without checking to see if the type in question does reference equality.

YingP99:
@jaredpar you are right. This is not meant for product code, but for converting some Assert in xUnit tests, and I don't want to get it complicated by involving semantic analysis. I put here because other test related rules such as convert MSTest to xUnit tests are also in this project.
Where is or will be another project for non-coding guideline rules?

jaredpar:
@YingP99 

I think this repo is the right place for such conversion tools, we just need to create a different application for it.  I'm actually doing that right now for the rest of the xUnit conversion code.  I'm not sure if this rule should or shouldn't go into that one.

YingP99:
@jaredpar  It definitely should go with the xUnit test convert rule. Can I merge this pull request first, then you move it? I'll be on vacation next week (leave on Sunday), so I'd like to get it in today.

jaredpar:
@YingP99 sure, that's fine.  I can just pull it into my branch later today.

YingP99:
@jaredpar merged. Please move it to another project.
If you have time and agree with @ellismg, please address his comments as well (I'm not disagree with him). Thx.

PR #64, opened 2015-02-15
Add rules for formatting attributes

terrajobst:
This adds two additional rules for dealing with attributes:
### Removing superfluous parentheses

It's conventional to omit parentheses when the attribute doesn't take any arguments, for example:

``` CSharp
[target: SomeAtt()]
```

becomes

``` CSharp
[target: SomeAtt]
```
### Separating attribute lists

Most folks only use a single attribute in an attribute list, although the syntax supports multiple attributes. In most cases, putting attributes in separate lists is beneficial because it improves readability and allows better diffs.

This changes code like this:

``` CSharp
[Serializable, Category("Catagory"), DefaultValue(1)]
```

to 

``` CSharp
[Serializable]
[Category("Catagory")]
[DefaultValue(1)]
```

The only exception are attributes for parameters. In interop code, it's quite common to mark parameters with `[In]` and `[Out]`. Since the attributes are usually written inline, i.e. not on separate lines, it's actually beneficial to no split them into separate lists.

This means that code like this is left as-is:

``` CSharp
void NativeFun([In, Out] int* target)
```

### DIFF
--- a/src/Microsoft.DotNet.CodeFormatting/Rules/AttributeNoParenthesesRule.cs
+++ b/src/Microsoft.DotNet.CodeFormatting/Rules/AttributeNoParenthesesRule.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Linq;
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Microsoft.DotNet.CodeFormatting.Rules
+{
+    [SyntaxRuleOrder(SyntaxRuleOrder.AttributeNoParenthesesRule)]
+    internal sealed class AttributeNoParenthesesRule : ISyntaxFormattingRule
+    {
+        public SyntaxNode Process(SyntaxNode syntaxRoot)
+        {
+            var attributes = syntaxRoot.DescendantNodes()
+                                       .OfType<AttributeSyntax>()
+                                       .Where(a => a.ArgumentList != null &&
+                                                   a.ArgumentList.Arguments.Count == 0 &&
+                                                   (!a.ArgumentList.OpenParenToken.IsMissing || !a.ArgumentList.CloseParenToken.IsMissing));
+
+            return syntaxRoot.ReplaceNodes(attributes, (a, n) => a.WithArgumentList(null));
+        }
+    }
+}
--- a/src/Microsoft.DotNet.CodeFormatting/Rules/AttributeSeparateListsRule.cs
+++ b/src/Microsoft.DotNet.CodeFormatting/Rules/AttributeSeparateListsRule.cs
@@ -1,0 +1,84 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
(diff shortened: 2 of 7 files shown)


terrajobst:
@jaredpar

src/Microsoft.DotNet.CodeFormatting/Rules/AttributeNoParenthesesRule.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        public SyntaxNode Process(SyntaxNode syntaxRoot)
+        {
+            var attributes = syntaxRoot.DescendantNodes()
+                                       .OfType<AttributeSyntax>()
+                                       .Where(a => a.ArgumentList != null &&
+                                                   a.ArgumentList.Arguments.Count == 0 &&
+                                                   (!a.ArgumentList.OpenParenToken.IsMissing || !a.ArgumentList.CloseParenToken.IsMissing));
+
+            return syntaxRoot.ReplaceNodes(attributes, (a, n) => a.WithArgumentList(null));
```
jaredpar: Consider the following case:

``` csharp
[Fact(/* Comment */)]
```

This code would strip away that comment.  Was that an intentional decision?  

In general I don't think we should be removing comments in a code formatter.  However I just can't see how such a comment would add value to the system so in this narrow case it seems okay.

jaredpar: Also I would use 

```
a.WithArgumentList(SyntaxFactory.ArgumentList())
```

I couldn't find any reference as to whether or not that property guaranteed non-nullness but it seems to for well formed code.

src/Microsoft.DotNet.CodeFormatting/Rules/AttributeSeparateListsRule.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Linq;
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Microsoft.DotNet.CodeFormatting.Rules
+{
+    [SyntaxRuleOrder(SyntaxRuleOrder.AttributeSeparateListsRule)]
+    internal sealed class AttributeSeparateListsRule : ISyntaxFormattingRule
```
jaredpar: Can you add a description of what this rule is trying to accomplish

src/Microsoft.DotNet.CodeFormatting/Rules/AttributeSeparateListsRule.cs
```diff
@@ -0,0 +1,84 @@
...
+                    var leadingTrivia = attributeList.GetLeadingTrivia()
+                                            .Concat(attributeList.OpenBracketToken.TrailingTrivia)
+                                            .Concat(result[firstIndex].GetLeadingTrivia());
+
+                    var trailingTrivia = result[lastIndex].GetTrailingTrivia()
+                                            .Concat(attributeList.CloseBracketToken.LeadingTrivia)
+                                            .Concat(attributeList.GetTrailingTrivia());
+
+                    result[firstIndex] = result[firstIndex].WithLeadingTrivia(leadingTrivia);
+                    result[lastIndex] = result[lastIndex].WithTrailingTrivia(trailingTrivia);
```
jaredpar: I've been staring at this for ~5 minutes now thinking about the trivia implications.  Consider the following:

``` csharp
[Attr1,
    Attr2]
```

I'm fairly certain this code would end up retaining the extra new line and spacing with the above logic.  

I'm struggling to think of a better approach though.  I considered recommending that you strip everything but comments and structured trivia and add that back in but that has some pathological cases to consider.  

Trivia isn't trivial :(

jaredpar:
:+1:

Gave some feedback on a couple of small issues to consider.

terrajobst:
Sorry, but I can't change the code anymore as I have the perfect number of lines being added! :-)

Just kidding, I'll take a look.

PR #76, opened 2015-02-23
Fix multiline copyright header rule

krwq:
Fix a problem where some lines of multiline header are repeated during the formatting already formmated code.

I.e:
```
// test1
// test2
```

was formatted to:
```
// test1
// test2

// test1
// test2
```

When using real headers some of those lines are removed as they are treated as "old" header. I.e:
```
// Copyright blah
// License X
```

was being formatted to:
```
// Copyright blah
// License X

// License blah
```

### DIFF
--- a/src/Microsoft.DotNet.CodeFormatting.Tests/Rules/CopyrightHeaderRuleTests.cs
+++ b/src/Microsoft.DotNet.CodeFormatting.Tests/Rules/CopyrightHeaderRuleTests.cs
@@ -37,6 +37,25 @@ class C
         }
 
         [Fact]
+        public void CSharpSimpleMultiline()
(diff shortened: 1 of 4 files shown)


krwq:
I just found an issue with that, doesn't work correctly with:
```
// Copyright blah
// License X

// My Comment
```

krwq:
Above is fixed in the updated PR.

krwq:
Above fixes the issue where there are whitespaces after the comment in the header (really rare but could happen).

krwq:
Hi @jaredpar, could you take a look at this PR?

jaredpar:
:+1: 

Thanks for fixing that.

PR #81, opened 2015-02-27
Fix a couple of newline issues

jaredpar:
The two HasNewLineBefore rules had a couple of small issues that needed
to be addressed.

1. They were unconditionally using CRLF which could possibly lead to
mixed line endings in files.  The code was changed to prefer existing
endings instead.
2. They were formatting around #if blocks which is suspect and needs to
be disabled.
3.  Produced new SyntaxNode values on trees that already had the correct
formatting.  This was potentially changing line endings and made it
difficult to find noisy rules in the formatter.

After looking at the logic of these rules they appear to really just be
the same rule run against two different nodes.  Hence I merged them into
a single rule to make it easier to share code.


jaredpar:
Merging this new because I found a more serious issue I need to fix that depends on this change.  Even though I'm merging this now I will respond / update code based on any feedback left on this PR.

PR #85, opened 2015-02-27
Add deadregions tool

nslottow:
Add deadregions tool I've been working on for analyzing and cleaning up unnecessary ("dead") conditional regions in C# code.


terrajobst:
@jaredpar, do you want to take a look?

jaredpar:
Yes.  Will look.

build.props
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+	  <OutDir Condition="'$(OutDir)' == ''">$(MsBuildThisFileDirectory)bin\</OutDir>
```
jaredpar: Can be shifted left

src/CodeFormatter/CodeFormatter.csproj
```diff
@@ -1,35 +1,17 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.props))\build.props" />
```
jaredpar: Is there a reason to not simplify this to:

``` xml
<Import Project="..\build.props />
```

src/CodeFormatter/CodeFormatter.csproj
```diff
@@ -1,35 +1,17 @@
...
     <ProjectGuid>{B0E1A988-F762-459D-AD0D-56A3CF4FFF3F}</ProjectGuid>
     <OutputType>Exe</OutputType>
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>CodeFormatter</RootNamespace>
     <AssemblyName>CodeFormatter</AssemblyName>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
-    <FileAlignment>512</FileAlignment>
   </PropertyGroup>
+  <!-- Default configurations to help VS understand -->
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
```
jaredpar: Nit: could simplify to a one liner `<PropertyGroup ... />`

src/DeadRegions/Program.cs
```diff
@@ -0,0 +1,290 @@
...
+        private static AnalysisEngine s_engine;
+
+        private static IList<string> s_filePaths;
+        private static List<string> s_ignoredSymbols = new List<string>();
+        private static List<string> s_definedSymbols = new List<string>();
+        private static List<string> s_disabledSymbols = new List<string>();
+        private static List<IEnumerable<string>> s_symbolConfigurations = new List<IEnumerable<string>>();
+        private static Tristate s_undefinedSymbolValue = Tristate.Varying;
+
+        private static bool s_printDisabled;
```
jaredpar: Mutable statics ...  why not put all of this global state into a Options / State type and pass it to the appropriate code paths?

src/Microsoft.DotNet.DeadRegionAnalysis.Tests/PreprocessorExpressionEvaluatorTests.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            Assert.Equal(Tristate.False, EvaluateExpression("true", evaluator));
+            Assert.Equal(Tristate.True, EvaluateExpression("false", evaluator));
+        }
+
+        private static readonly CSharpParseOptions s_expressionParseOptions = new CSharpParseOptions(
+            documentationMode: DocumentationMode.None,
+            kind: SourceCodeKind.Interactive);
+
+        private static readonly PreprocessorExpressionEvaluator s_evaluator = new PreprocessorExpressionEvaluator(
```
jaredpar: Why static here?  This is a mutable type from what I can tell.  I would make it instance state.  Making it static means we have to proactively disable xUnit features like running in parallel.

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.Options.cs
```diff
@@ -0,0 +1,200 @@
...
+                IEnumerable<string> sourcePaths = null,
+                IEnumerable<IEnumerable<string>> symbolConfigurations = null,
+                IEnumerable<string> alwaysIgnoredSymbols = null,
+                IEnumerable<string> alwaysDefinedSymbols = null,
+                IEnumerable<string> alwaysDisabledSymbols = null,
+                Tristate undefinedSymbolValue = default(Tristate))
+            {
+                if (projectPaths != null)
+                {
+                    projects = Task.WhenAll(from path in projectPaths select MSBuildWorkspace.Create().OpenProjectAsync(path, CancellationToken.None)).Result;
```
jaredpar: Why not make using `await` here instead?

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.Options.cs
```diff
@@ -0,0 +1,200 @@
...
+                        .CurrentSolution
+                        .AddProject(projectId, "AnalysisProject", "AnalysisProject", LanguageNames.CSharp);
+
+                    foreach (var path in sourcePaths)
+                    {
+                        var documentId = DocumentId.CreateNewId(projectId);
+                        solution = solution.AddDocument(
+                            documentId,
+                            Path.GetFileName(path),
+                            new FileTextLoader(path, Encoding.UTF8));
```
jaredpar: Using an explicit encoding here will break files with a non-UTF8 encoding.  We should just load with encoding detection enabled.

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.Options.cs
```diff
@@ -0,0 +1,200 @@
...
+
+            internal CompositePreprocessorExpressionEvaluator GetPreprocessorExpressionEvaluator()
+            {
+                var evaluators = _symbolConfigurations.Select(config => new PreprocessorExpressionEvaluator(config, _undefinedSymbolValue));
+                return new CompositePreprocessorExpressionEvaluator(evaluators);
+            }
+
+            internal PreprocessorSymbolTracker GetPreprocessorSymbolTracker()
+            {
+                var specifiedSymbols = new HashSet<string>();
```
jaredpar: Nit: I would specify `StringComparer.Oridinal` just to be explicit here.

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.Options.cs
```diff
@@ -0,0 +1,200 @@
...
+namespace Microsoft.DotNet.DeadRegionAnalysis
+{
+    public partial class AnalysisEngine
+    {
+        public class Options
+        {
+            private IEnumerable<IReadOnlyDictionary<string, Tristate>> _symbolConfigurations;
+            private Tristate _undefinedSymbolValue;
+
+            public IEnumerable<Document> Documents { get; private set; }
```
jaredpar: Enumerable suggests the values can change.  If they can't we should use `ImmutableArray<Document>` instead.

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.Options.cs
```diff
@@ -0,0 +1,200 @@
...
+                        specifiedSymbols.Add(symbol);
+                    }
+                }
+
+                return new PreprocessorSymbolTracker(specifiedSymbols);
+            }
+
+            private static IEnumerable<Document> GetSharedDocuments(IEnumerable<Project> projects)
+            {
+                var it = projects.GetEnumerator();
```
jaredpar: Need to put this in a using block since `IEnumerator<T>` is disposable.

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.Options.cs
```diff
@@ -0,0 +1,200 @@
...
+                    return Enumerable.Empty<Document>();
+                }
+
+                var filePaths = it.Current.Documents.Select(d => d.FilePath);
+
+                while (it.MoveNext())
+                {
+                    filePaths = filePaths.Intersect(
+                        it.Current.Documents.Select(d => d.FilePath),
+                        StringComparer.InvariantCultureIgnoreCase);
```
jaredpar: File Path comparison should be `OrdinalIgnoreCase`

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.Options.cs
```diff
@@ -0,0 +1,200 @@
...
+                }
+
+                var filePathSet = new HashSet<string>(filePaths);
```
jaredpar: Need to pass `StringComparer.OrdinalIgnoreCase` here as well.

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.Options.cs
```diff
@@ -0,0 +1,200 @@
...
+                    foreach (string symbol in config.Keys)
+                    {
+                        specifiedSymbols.Add(symbol);
+                    }
+                }
+
+                return new PreprocessorSymbolTracker(specifiedSymbols);
+            }
+
+            private static IEnumerable<Document> GetSharedDocuments(IEnumerable<Project> projects)
```
jaredpar: I think the logic of this method could be simplified to:

``` csharp
using (var it = projects.GetEnumerator())
{
    if (!it.MoveNext())
    {
        return Enumerable.Empty<Document>()
    }

    var first = it.Current;
    var filePaths = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
    while (it.MoveNext())
    {
        foreach (var doc in it.Documents)
        {
            filePaths.Add(doc.FilePath);
        }
    }

    return first.Documents.Where(d => filePaths.Contains(d.FilePath));
}
```

This would avoid all of the internal allocations in `Intersect`

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.RegionRemoval.cs
```diff
@@ -0,0 +1,196 @@
...
+                var changesString = new StringBuilder();
+                var syntaxTree = info.Document.GetSyntaxTreeAsync(cancellationToken).Result;
+
+                foreach (var change in changes)
+                {
+                    var lineSpan = Location.Create(syntaxTree, change.Span).GetLineSpan();
+                    changesString.AppendFormat("({0}-{1}): {2}", lineSpan.StartLinePosition.Line, lineSpan.EndLinePosition.Line, newText.GetSubText(change.Span).ToString());
+                }
+
+                Console.WriteLine(string.Format("Failed to remove regions from document '{0}':{1}{2}", info.Document.FilePath, Environment.NewLine, changesString.ToString()));
```
jaredpar: If this is meant to be re-hostable we should avoid the console here and instead abstract it out to an interface.

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.RegionRemoval.cs
```diff
@@ -0,0 +1,196 @@
...
+            {
+                var region = chain.Regions[i];
+                if (region.State != Tristate.Varying)
+                {
+                    var startDirective = region.StartDirective;
+                    var endDirective = region.EndDirective;
+                    string endDirectiveReplacementText = string.Empty;
+
+                    // Remove the start directive
+                    changes.Add(new TextChange(new TextSpan(region.SpanStart, region.StartDirective.FullSpan.End - region.SpanStart), string.Empty));
```
jaredpar: Why can't we use `region.StartDirective.FullSpan` for `TextSpan`?

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.RegionRemoval.cs
```diff
@@ -0,0 +1,196 @@
...
+            }
+            else
+            {
+                return endDirective.ToFullString();
+            }
+        }
+
+        private static int CompareTextChanges(TextChange x, TextChange y)
+        {
+            int result = x.Span.End - y.Span.End;
```
jaredpar: Could simplify to `return x.Span.CompareTo(y.Span)`

src/Microsoft.DotNet.DeadRegionAnalysis/ConditionalRegion.cs
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        public override bool Equals(object obj)
+        {
+            return Equals(obj as ConditionalRegion);
+        }
+
+        public override int GetHashCode()
+        {
+            return base.GetHashCode();
```
jaredpar: That can't be used here.  Two values which are `Equals` must also have the same hash code.  The base implementation of `GetHashCode` won't satisfy this requirement given the implementation of `Equals`

src/Microsoft.DotNet.DeadRegionAnalysis/ConditionalRegion.cs
```diff
@@ -0,0 +1,125 @@
...
+            }
+
+            return start;
+        }
+
+        public int CompareTo(ConditionalRegion other)
+        {
+            if (other == null)
+            {
+                return 1;
```
jaredpar: This is saying that if a collection of `ConditionalRegion` is sorted the `null` should come first.  Generally they are placed last.

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.RegionRemoval.cs
nslottow: region.StartDirective.FullSpan does not include any whitespace trivia preceding the DirectiveTriviaSyntax.  That's the reason for the region.SpanStart property.

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.Options.cs
nslottow: Sounds good.

As a note to myself: I don't think I've documented the fact that we take the intersection of documents in the specified projects when multiple projects are passed on the command line.  Originally, I passed two projects which have different build configurations, but many shared files on the command line, and intersected the state of all the conditional regions in each project.

Since updating the algorithm, the build configurations specified by the projects themselves are ignored because whichever configurations msbuild deduces from the project files is dependent on the environment.  It's clearer to specify each configuration explicitly on the command line, an easy way to get each configuration being to look at the csc commands run by msbuild.

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.RegionRemoval.cs
jaredpar: Interesting.  I didn't realize that.

src/CodeFormatter/CodeFormatter.csproj
nslottow: This would actually be `..\..\build.props` for all the projects we have right now.  I basically just did this so I could put the build.props at the top-level without having to specify exactly where the top-level is (in case things move it's one less thing to change for each project).  This is what we do in the corefx repo to find `dirs.props` and `dirs.targets`, but since the build here is not as complex this is probably not necessary.   I'll change this to `..\..\build.props`.

src/Microsoft.DotNet.DeadRegionAnalysis.Tests/PreprocessorExpressionEvaluatorTests.cs
nslottow: Does xUnit parallel test execution actually get automatically disabled if we have static fields in test classes?  The PreprocessorExpressionEvaluator and PreprocessorExpressionSimplifier do not actually have any mutable state.  I guess technically for me to claim that I would have to be copying the dictionaries passed into the constructors of these classes.  Is that what you're referring to when you say this type is mutable?

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.Options.cs
nslottow: Correction: I thought I had changed this, but it turns out I am not ignoring the symbol configuration specified in the project files in the AnalysisEngine.FromProjects() method (currently only used by tests).  I am currently using it in addition to any configurations specified with the `/config` option.  Given my explanation above though, I think it makes sense never to use the symbol configuration deduced by simply loading the project file in the current environment.

src/DeadRegions/Program.cs
nslottow: Sure.

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.Options.cs
nslottow: `await` can't be used inside a constructor, but I could pull out the project loading into AnalysisEngine.FromFilePaths() and make that async.

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.Options.cs
nslottow: Will fix.

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.Options.cs
nslottow: Do you have a suggestion for how to do this?

src/CodeFormatter/App.config
```diff
@@ -1,42 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<configuration>
-    <startup> 
-        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5" />
-    </startup>
-  <runtime>
-    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
```
nslottow: Are these binding redirects really necessary? Everything works fine for me without them, and I've tried this both with VS2013 and VS2015 CTP.

src/Microsoft.DotNet.CodeFormatting.Tests/app.config
```diff
@@ -1,15 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<configuration>
-  <runtime>
-    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
```
nslottow: This file is unnecessary: App.config files don't do anything for library projects by default.

src/Microsoft.DotNet.CodeFormatting/app.config
```diff
@@ -1,15 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<configuration>
-  <runtime>
```
nslottow: Are these binding redirects really necessary?

src/Microsoft.DotNet.DeadRegionAnalysis.Tests/PreprocessorExpressionEvaluatorTests.cs
jaredpar: It depends on the host but by default it is okay to parallelize tests.  

Hmm, I just assumed those types were mutable hence would be a problem as a static.  If they are truly immutable it's fine to keep them there.  But still I would juts make them member fields and use them that way.  There is no real downside to them being instance fields.

src/CodeFormatter/App.config
jaredpar: Please add this file back.

These bindings are needed on machines which have older versions of MSBuild / Roslyn installed into the GAC.

nslottow: Will do.  I'll also add a similar App.config for the DeadRegions executable.

src/Microsoft.DotNet.DeadRegionAnalysis/AnalysisEngine.Options.cs
jaredpar: Actually this API behaves already does this.  The `Encoding` argument isn't dictating an encoding but instead specifying a backup one.  Can you change this be passed as a named argument?  It helps remove this confusion.

nslottow: Sure.

src/Microsoft.DotNet.DeadRegionAnalysis.Tests/PreprocessorExpressionEvaluatorTests.cs
nslottow: The only downside is additional allocations per test method.  I'll assume we don't care about that though. Fixing.

nslottow: On second thought, I don't see a great reason to change this.  It's a common pattern for us to declare const or static readonly test objects.  For example: [ImmutableArrayTests.cs](https://github.com/dotnet/corefx/blob/master/src/System.Collections.Immutable/tests/ImmutableArrayTest.cs).

src/CodeFormatter/App.config
nslottow: And the XunitConverter executable.

jaredpar:
:+1:

nslottow:
Thanks for the review.

PR #91, opened 2015-03-06
Add support for cleaning up files without a solution or project.

### LINKED ISSUE #89: Add support for formatting C# that doesn't have a solution or project file.
Having a solution or msbuild project should not be a requirement for being able to run the codeformatter.

In other words: instead of passing: 
    <project or solution> 
to the tool someone should be able to alternatively pass:
    [/r:<reference>] [/cs:<sourcefile>]

The workspace that interanlly gets created would then not be an MsBuildWorkspace, but an AdHocWorkspace that is populated with a single project that contains the references and sources.

dannyvv:
The passed file, can now not only be an msbuild project or a solution file. But also an rsp file.
The rsp file is the same file format as the commandline arguments that gets passed to csc.exe

Closes #89

### DIFF
--- a/src/CodeFormatter/Program.cs
+++ b/src/CodeFormatter/Program.cs
@@ -2,13 +2,13 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.MSBuild;
 using Microsoft.DotNet.CodeFormatting;
 
@@ -19,18 +19,21 @@ namespace CodeFormatter
         private const string FileSwitch = "/file:";
         private const string ConfigSwitch = "/c:";
         private const string CopyrightSwitch = "/copyright:";
+        private const string LanguageSwitch = "/lang:";
 
         private static int Main(string[] args)
         {
             if (args.Length < 1)
             {
-                Console.WriteLine("CodeFormatter <project or solution> [/file:<filename>] [/nocopyright] [/nounicode] [/tables] [/c:<config1,config2> [/copyright:file] [/verbose]");
+                Console.WriteLine("CodeFormatter <project, solution or responsefile> [/file:<filename>] [/nocopyright] [/nounicode] [/tables] [/c:<config1,config2> [/copyright:file] [/lang:<language>] [/verbose]");
                 Console.WriteLine("    <filename>   - Only apply changes to files with specified name.");
                 Console.WriteLine("    <configs>    - Additional preprocessor configurations the formatter");
                 Console.WriteLine("                   should run under.");
                 Console.WriteLine("    <copyright>  - Specifies file containing copyright header.");
                 Console.WriteLine("                   Use ConvertTests to convert MSTest tests to xUnit.");
                 Console.WriteLine("    <tables>     - Let tables opt out of formatting by defining DOTNET_FORMATTER");
+                Console.WriteLine("    <language>   - Specifies the language to use when a responsefile is specified.");
+                Console.WriteLine("                   i.e. 'C#', 'Visual Basic', ... (default: 'C#')");
                 Console.WriteLine("    <verbose>    - Verbose output");
                 Console.WriteLine("    <nounicode>  - Do not convert unicode strings to escape sequences");
                 Console.WriteLine("    <nocopyright>- Do not update the copyright message.");
@@ -40,7 +43,7 @@ namespace CodeFormatter
             var projectOrSolutionPath = args[0];
             if (!File.Exists(projectOrSolutionPath))
             {
-                Console.Error.WriteLine("Project or solution {0} doesn't exist.", projectOrSolutionPath);
+                Console.Error.WriteLine("Project, solution or response file {0} doesn't exist.", projectOrSolutionPath);
                 return -1;
             }
 
@@ -48,6 +51,7 @@ namespace CodeFormatter
             var ruleTypeBuilder = ImmutableArray.CreateBuilder<string>();
             var configBuilder = ImmutableArray.CreateBuilder<string[]>();
             var copyrightHeader = FormattingConstants.DefaultCopyrightHeader;
+            var language = LanguageNames.CSharp;
             var convertUnicode = true;
             var allowTables = false;
             var verbose = false;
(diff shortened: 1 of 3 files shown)


ellismg:
LGTM.  @jaredpar do you have any feedback?  I think before merging we should rebase this down to one change without any merge commits.

src/CodeFormatter/Program.cs
```diff
@@ -159,18 +168,32 @@ private static int Main(string[] args)
...
 
-            Console.WriteLine(Path.GetFileName(projectOrSolutionPath));
-            string extension = Path.GetExtension(projectOrSolutionPath);
-            if (StringComparer.OrdinalIgnoreCase.Equals(extension, ".sln"))
+            Console.WriteLine(Path.GetFileName(projectSolutionOrRspPath));
+            string extension = Path.GetExtension(projectSolutionOrRspPath);
+            if (StringComparer.OrdinalIgnoreCase.Equals(extension, ".rsp"))
+            {
+                using (var workspace = ResponseFileWorkspace.Create())
+                {
```
jaredpar: Should add the following line:

``` csharp
workspace.LoadMetadataForReferencedProjects = true;
```

This can help increase performance by loading referenced projects as metedata vs. source.  Shows up mostly with larger solutions (aka Roslyn).

src/Microsoft.DotNet.CodeFormatting/ResponseFileWorkspace.cs
```diff
@@ -0,0 +1,75 @@
...
+        protected override void ApplyDocumentTextChanged(DocumentId documentId, SourceText text)
+        {
+            var document = this.CurrentSolution.GetDocument(documentId);
+            if (document != null)
+            {
+                try
+                {
+                    var filePath = document.FilePath;
+                    Directory.CreateDirectory(Path.GetDirectoryName(filePath));
+                    using (var writer = new StreamWriter(filePath, append: false, encoding: Encoding.UTF8))
```
jaredpar: Need to be *very* careful with encoding here.  Can't assume that the files are all in UTF8.

The best way to proceed here is to grab the encoding of the original `Document` and use that.

src/Microsoft.DotNet.CodeFormatting/ResponseFileWorkspace.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        protected override void ApplyDocumentTextChanged(DocumentId documentId, SourceText text)
+        {
+            var document = this.CurrentSolution.GetDocument(documentId);
+            if (document != null)
+            {
+                try
+                {
+                    var filePath = document.FilePath;
+                    Directory.CreateDirectory(Path.GetDirectoryName(filePath));
```
jaredpar: This seems unnecessary.

src/Microsoft.DotNet.CodeFormatting/ResponseFileWorkspace.cs
dannyvv: All the code for this was private in Roslyn. Thought I'd keep it simple. I'll copy the Roslyn code and keep a note when it is public there to use the public version.

src/CodeFormatter/Program.cs
dannyvv: done

src/Microsoft.DotNet.CodeFormatting/ResponseFileWorkspace.cs
dannyvv: Done

src/Microsoft.DotNet.CodeFormatting/ResponseFileWorkspace.cs
dannyvv: Default workspaces can't write files, Hence why I had to create a ResponseFileWorksapce and couldn't use the AdHoc workspace.
I looked at the code of MsBuildWorkspace and this is the way the MsBuild one writes back. 
Do you have a more concrete way to write back?

dannyvv: MsBuild workspace creates the directory. I presume this was in case documents get added to the workspace. 
Let me know and I'll yank it. In my opinion the Filesystem should have created the parent dirs. if they didn't exist, or the .Net stream wrappers should have done it.

jaredpar: What's confusing me is the code formatter shouldn't be creating any new files or renaming old ones.  Hence I'm unsure how we could ever get into a state where the `Document` being written didn't have a directory yet.

src/Microsoft.DotNet.CodeFormatting/ResponseFileWorkspace.cs
```diff
@@ -0,0 +1,76 @@
...
+                    using (var writer = new StreamWriter(filePath, append: false, encoding: text.Encoding ?? Encoding.UTF8))
```
jaredpar: In the case the encoding doesn't exist we should use the `UTF8` encoding which doesn't emit a BOM.  Same type, just pass false to the UTF8 constructor.  

There is no right answer here because it's encoding.  But this is what we've found to be the most popular form in our uses.

src/Microsoft.DotNet.CodeFormatting/ResponseFileWorkspace.cs
dannyvv: K I'll yank it...

dannyvv: done

src/Microsoft.DotNet.CodeFormatting/ResponseFileWorkspace.cs
dannyvv: Done, but you should probably file a bug on Roslyn because the MsBuildWorkspace emits with BOM. Other side-note: I've seen some of the git tools infer the wrong encoding if the BOM isn't there. Causing big issues with the windows vs Unix line endings mess.

jaredpar:
@dannyvv where are you seeing MSBuildWorkspace emit with a BOM?  That is definitely a bug if it is happening.  

FWIW: It doesn't emit BOMs on any files when we run it against Roslyn.

dannyvv:
@jaredpar : http://index/#Microsoft.CodeAnalysis.Workspaces.Desktop/Workspace/MSBuild/MSBuildWorkspace.cs,751dabb506683b7c
Line 972.

jaredpar:
@dannyvv that feels like a bug.  Filed the following to track it.

https://github.com/dotnet/roslyn/issues/1146

jaredpar:
Thanks @dannyvv !!!

PR #93, opened 2015-03-07
Add an option to run CodeFormatter on all projects/solutions within a directory.

AlekseyTs:
Allow to specify a directory instead of a project/solution.

### DIFF
--- a/src/CodeFormatter/Program.cs
+++ b/src/CodeFormatter/Program.cs
@@ -19,12 +19,13 @@ namespace CodeFormatter
         private const string FileSwitch = "/file:";
         private const string ConfigSwitch = "/c:";
         private const string CopyrightSwitch = "/copyright:";
+        private const string SearchPatternSwitch = "/searchPattern:";
 
         private static int Main(string[] args)
         {
             if (args.Length < 1)
             {
-                Console.WriteLine("CodeFormatter <project or solution> [/file:<filename>] [/nocopyright] [/nounicode] [/tables] [/c:<config1,config2> [/copyright:file] [/verbose]");
+                Console.WriteLine("CodeFormatter <project or solution or directory> [/file:<filename>] [/nocopyright] [/nounicode] [/tables] [/c:<config1,config2> [/copyright:file] [/recursive] [/searchPattern:<searchPattern>] [/verbose]");
                 Console.WriteLine("    <filename>   - Only apply changes to files with specified name.");
                 Console.WriteLine("    <configs>    - Additional preprocessor configurations the formatter");
                 Console.WriteLine("                   should run under.");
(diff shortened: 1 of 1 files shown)


src/CodeFormatter/Program.cs
```diff
@@ -34,14 +35,30 @@ private static int Main(string[] args)
...
                 Console.WriteLine("    <nounicode>  - Do not convert unicode strings to escape sequences");
                 Console.WriteLine("    <nocopyright>- Do not update the copyright message.");
+                Console.WriteLine("    <searchPattern>- Pattern to look for projects/solutions within the folder. Default '*.??proj'.");
                 return -1;
             }
 
-            var projectOrSolutionPath = args[0];
-            if (!File.Exists(projectOrSolutionPath))
+            var path = args[0];
+            string[] projectOrSolutionPath;
```
jaredpar: We should rename this type to reflect that it is a collection.

src/CodeFormatter/Program.cs
```diff
@@ -111,17 +152,32 @@ private static int Main(string[] args)
...
+                }
+
+                if (projectOrSolutionPath.Length > 1)
+                {
+                    Console.WriteLine("Completed formatting {0} projects/solutions.", projectOrSolutionPath.Length);
+                }
+                else
+                {
+                    Console.WriteLine("Completed formatting.");
+                }
```
jaredpar: We should move the status line up to be inside the loop.  That way the user gets feedback for every project / solution that is formatted.

src/CodeFormatter/Program.cs
AlekseyTs: Did you mean to rename the local?

src/CodeFormatter/Program.cs
AlekseyTs: User already gets feedback for every project / solution that is formatted. The output looks like this:
--------------------
Proj1.csproj
        Syntax Pass
        Local Semantic Pass
        Global Semantic Pass
Total time 00:00:00.0798914
Proj2.csproj
        Syntax Pass
        Local Semantic Pass
        Global Semantic Pass
Total time 00:00:00.0225966
Completed formatting 2 projects/solutions.

jaredpar: Ah yes, I forgot about that.

jaredpar:
I'm curious about why this approach vs. just using the shell to feed projects to the tool?  Usually I just do something like the following to run code formatter on large directory trees.  

```
gci -re -in *proj | %{ codeformatter $_ } 
```

src/CodeFormatter/Program.cs
jaredpar: @AlekseyTs Yes.

AlekseyTs:
I prefer tool handling this scenario and also giving me the tally at the end. Also, not everyone is proficient with shell commands.

VSadov:
is this still active?

Issue #96, opened 2015-03-11
Issue when existing header begins with copyright header


AndyGerlicher:
I think I found an issue in the logic that determines if a file already has the copyright header or not. Right now it looks at the leading trivia and if that sequence is equal to the copyright, it returns true. I think it should be if the sequence starts with, not equals.

For my particular case, I first ran the tool the first time quite a while back (before it even went to GitHub), so I don't know if the intended output has changed since then. But when I ran it again recently I got **many** changes with the headers clearly messed up. Here's an example:

Current file header (output from previous run):
```
// Copyright (c) Microsoft. All rights reserved.
// Licensed under the MIT license. See LICENSE file in the project root for full license information.
//-----------------------------------------------------------------------
// </copyright>
// <summary>Utility methods for classifying and handling exceptions.</summary>
//-----------------------------------------------------------------------
```

Expected: No change. Obviously that `// </copyright>` line shouldn't be there, but that's just an artifact from the first time I ran it when it removed most of the existing copyright header. I'll have to clean those up manually.

Actual:
```
// Copyright (c) Microsoft. All rights reserved.
// Licensed under the MIT license. See LICENSE file in the project root for full license information.

// Licensed under the MIT license. See LICENSE file in the project root for full license information.
//-----------------------------------------------------------------------
// </copyright>
// <summary>Utility methods for classifying and handling exceptions.</summary>
//-----------------------------------------------------------------------
```

I was able to fix the issue by changing HasCopyrightHeader to check if the SequenceStartsWith vs SequenceEquals.

PR #97, opened 2015-03-11
Fix issue if a file leading trivia starts with the copyright header

AndyGerlicher:
This should fix issues causing the file header to be messed up when the copyright header is already at the top of the file. See Issue #96 for details.


krwq:
LGTM :+1:

jaredpar:
:+1: 

Thanks for contributing that fix!

Issue #78, opened 2015-02-24
Copyright should be written last at the top of file to avoid incorrect placement.


jterry75:
Consider the following example:

```C#
namespace Test
{
    using System;
    
    class TestClass
    {
        public static void Main()
        {
            Console.WriteLine("Test");
        }
    }
}
```

After formatting:

```C#
using System;
// Copyright (c) Microsoft. All rights reserved.

namespace Test
{
    internal class TestClass
    {
        public static void Main()
        {
            Console.WriteLine("Test");
        }
     }
}
```

Because the using statement is internal to the namespace the copyright is placed in the wrong order.

*The workaround is to first format with /nocopyright. Then reformat with /copyright:file.txt

Thanks,
Justin

jterry75:
Sorry still learning* This is open.

PR #99, opened 2015-03-23
Change CopyrightHeaderRule order to run later

AndyGerlicher:
Fix for Issue #78

When moving the using block outside of a namespace declaration, the copyright header would end up lower in the file. On the next run, it would add another one. This change just changes the copyright rule to run later.


jterry75:
Can we merge this change?

PR #109, opened 2015-04-13
Readonly field formatter

### LINKED ISSUE #104: Private/Internal fields that are not written to should be marked readonly
Annotating fields as readonly if they’re internal/private and we can prove they’re only ever read from, aren’t mutating structs, etc.  There are potentially lots of fields across the codebase that could be made readonly, and while that doesn’t currently help with performance, it’s very helpful for readability, hygiene, etc.

ChadNedzlek:
Add the readonly keyword to fields that are provably never written
to. Only mark private and internal fields, and the following
are considered "writing" to a field:

- Assignments (_field = 5)
- Compound assignments (_field += 5)
- ref/out calls (Method(ref _field))
- ref of parent to extern method. While not technically required
  the fields can possibly change (even when marked readonly)
- internal fields that are exposed via InternalsVisibleTo an
  assembly that is not in the input project

Fix #104

### DIFF
--- a/src/Microsoft.DotNet.CodeFormatting.Tests/Microsoft.DotNet.CodeFormatting.Tests.csproj
+++ b/src/Microsoft.DotNet.CodeFormatting.Tests/Microsoft.DotNet.CodeFormatting.Tests.csproj
@@ -68,6 +68,7 @@
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\packages\System.Collections.Immutable.1.1.33-beta\lib\portable-net45+win8+wp8+wpa81\System.Collections.Immutable.dll</HintPath>
     </Reference>
+    <Reference Include="System.ComponentModel.Composition" />
     <Reference Include="System.Composition.AttributedModel, Version=1.0.27.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\packages\Microsoft.Composition.1.0.27\lib\portable-net45+win8+wp8+wpa81\System.Composition.AttributedModel.dll</HintPath>
@@ -113,6 +114,7 @@
     <Compile Include="Rules\NewLineAboveRuleTests.cs" />
     <Compile Include="Rules\PrivateFieldNamingRuleTests.cs" />
     <Compile Include="Rules\NonAsciiCharactersAreEscapedInLiteralsRuleTests.cs" />
+    <Compile Include="Rules\MarkReadonlyFieldTests.cs" />
     <Compile Include="Rules\UsingLocationRuleTests.cs" />
   </ItemGroup>
   <ItemGroup>
--- a/src/Microsoft.DotNet.CodeFormatting/NameHelper.cs
+++ b/src/Microsoft.DotNet.CodeFormatting/NameHelper.cs
@@ -1,0 +1,55 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.DotNet.CodeFormatting
+{
+    /// <summary>
+    /// Commonly used name functions
+    /// </summary>
+    internal static class NameHelper
+    {
+        /// <summary>
+        /// Get the full name of a symbol. So a field might look like
+        /// "OuterNamespace.Inner.ClassName.FieldName"
+        /// </summary>
+        /// <param name="symbol">symbol to get name of</param>
+        /// <returns>Full display name for a symbol</returns>
+        internal static string GetFullName(ISymbol symbol)
+        {
+            return GetFullName(symbol.ContainingType) + "." + symbol.Name;
+        }
+
+        /// <summary>
+        /// Get the full name of a type. i.e. "OuterNamespace.Inner.ClassName"
+        /// </summary>
+        /// <param name="type">type to get name of</param>
+        /// <returns>Full display name for a type</returns>
+        internal static string GetFullName(INamedTypeSymbol type)
+        {
+            if (type.ContainingType != null)
+            {
+                return GetFullName(type.ContainingType) + "+" + type.Name;
+            }
+
+            return GetFullName(type.ContainingNamespace) + "." + type.Name;
+        }
+
+        /// <summary>
+        /// Get the full name of a namespace. i.e. "OuterNamespace.Inner.ClassName"
+        /// </summary>
+        /// <param name="namespaceSymbol">namespace to get name of</param>
+        /// <returns>Full display name for a namespaceSymbol</returns>
+        internal static string GetFullName(INamespaceSymbol namespaceSymbol)
+        {
+            if (namespaceSymbol.ContainingNamespace != null &&
+                !namespaceSymbol.ContainingNamespace.IsGlobalNamespace)
+            {
+                return GetFullName(namespaceSymbol.ContainingNamespace) + "." + namespaceSymbol.Name;
+            }
+
+            return namespaceSymbol.Name;
+        }
+    }
+}
--- a/src/Microsoft.DotNet.CodeFormatting/Rules/MarkReadonlyFieldsRule.cs
+++ b/src/Microsoft.DotNet.CodeFormatting/Rules/MarkReadonlyFieldsRule.cs
@@ -1,0 +1,463 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Microsoft.DotNet.CodeFormatting.Rules
+{
+    /// <summary>
+    /// Mark any fields that can provably be marked as readonly.
+    /// </summary>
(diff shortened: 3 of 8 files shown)


bartonjs:
cc: @stephentoub, @jaredpar

jaredpar:
Another scenario we have to consider a write is when a method is invoked on a field which is a struct.  Making that field `readonly` would cause a behavior change because the method would be invoked on a copy of the struct.  In the case it was a mutating struct this would lead to a behavior change.

ChadNedzlek:
That's very interesting, and not something I knew before today. Shouldn't take more than an hour to add a test and check for that scenario.

ChadNedzlek:
Hrm... I can't figure out any safe way to exclude things like Int32.ToString().  Technically it's a behavioral change, since now the int will get copied, but that's a sad reason not to make something readonly, just because you read it's string value.

jaredpar:
@ChadNedzlek hmm, it seems all of the useful extension methods to check this are `internal`.  That's unfortunate.  This though is logically what you need.  Could just copy / paste it into the rule.

http://source.roslyn.io/#Microsoft.CodeAnalysis/SpecialTypeExtensions.cs,103

src/Microsoft.DotNet.CodeFormatting/Rules/MarkReadonlyFieldsRule.cs
```diff
@@ -0,0 +1,458 @@
...
+                }
+            }
+
+            private void CheckForFieldWrite(ExpressionSyntax node)
+            {
+                var fieldSymbol = _semanticModel.GetSymbolInfo(node).Symbol as IFieldSymbol;
+
+                if (fieldSymbol != null)
+                {
+                    if (IsInsideOwnConstructor(node, fieldSymbol.ContainingType))
```
CyrusNajmabadi: What if you're inside a lambda inside the constructor?  Is that legal?

CyrusNajmabadi:
Does this also handle fields that are MEF imported? These should not be marked readonly as MEF will need to assign to them.

ChadNedzlek:
I've updated ChadNedzlek/codeformatter/readonly-fields-working. It handles the method calls and the delegates inside of constructors.  Going to work on the MEF importing next.

ChadNedzlek:
This failure seems to have nothing to do with my change, and looks like a nuget failure to pull some of the packages.

bartonjs:
@dotnet-bot test this please

ChadNedzlek:
Thank you Jeremy.

I've addressed all the comments in this thread, I believe.

src/Microsoft.DotNet.CodeFormatting/NameHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        /// <summary>
+        /// Get the full name of a symbol. So a field might look like
+        /// "OuterNamespace.Inner.ClassName.FieldName"
+        /// </summary>
+        /// <param name="symbol">symbol to get name of</param>
+        /// <returns>Full display name for a symbol</returns>
+        internal static string GetFullName(ISymbol symbol)
+        {
+            return GetFullName(symbol.ContainingType) + "." + symbol.Name;
```
jaredpar: Don't believe this will work for nested types.  They should use `+` instead of `.`

src/Microsoft.DotNet.CodeFormatting/Rules/MarkReadonlyFieldsRule.cs
```diff
@@ -0,0 +1,527 @@
...
+                    }
+
+                    string assemblyName = assemblyNameArgument.Value as string;
+                    if (String.IsNullOrEmpty(assemblyName))
+                    {
+                        // The first argument wasn't a string, isn't really the correct type
+                        continue;
+                    }
+
+                    if (!solution.Projects.Any(p => ProjectHasName(p, assemblyName)))
```
jaredpar: I don't think we can make this check safely.  It would cause different behavior for the formatter when it was run on a project vs. a solution.  Generally the two should be equal.

jaredpar: I would just simplify to if it has internals visible to then it's visible.

src/Microsoft.DotNet.CodeFormatting/NameHelper.cs
ChadNedzlek: Indeed so. Even though it’s only sort of a helper function internally (these names are only ever used to track down a type for a check), I’m going to update it to use the +, since it’s correct.

src/Microsoft.DotNet.CodeFormatting.Tests/Microsoft.DotNet.CodeFormatting.Tests.csproj
```diff
@@ -68,6 +68,7 @@
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\packages\System.Collections.Immutable.1.1.33-beta\lib\portable-net45+win8+wp8+wpa81\System.Collections.Immutable.dll</HintPath>
     </Reference>
+    <Reference Include="System.ComponentModel.Composition" />
```
jaredpar: Most of your attribute checking seems to be by name vs. by type.  Given that is this reference needed anymore?

ChadNedzlek: It’s used because I need to be able to compile the test input code. I could try and go scrape the assembly off the computer running the tests somehow, but the easiest way (and most reliable I can figure out) is just to add a reference to the tests, and then use the loaded assembly as a MetadataReference. Without the assembly, the type won’t get loaded, then I can’t reliably get the full name of it.

src/Microsoft.DotNet.CodeFormatting/Rules/MarkReadonlyFieldsRule.cs
ChadNedzlek: I worry that anything that has unit tests is going to expose their internals to the test assembly, which would disable readonly for any of the internal fields.  Running on the solution that contained both of them would solve the problem.  Running on the solution will mark more things (correctly) as readonly than running it only on the project.

However, it looks like most of the framework doesn't actually expose many of it's internals, so I'll remove the solution level check for now.  I'll still have the commit in my fork if we want to add it back later.

src/Microsoft.DotNet.CodeFormatting/Rules/MarkReadonlyFieldsRule.cs
```diff
@@ -0,0 +1,527 @@
...
+        }
+
+        public async Task<Solution> ProcessAsync(
+            Document document,
+            SyntaxNode syntaxRoot,
+            CancellationToken cancellationToken)
+        {
+            if (_unwrittenWritableFields == null)
+            {
+                using (await SemaphoreLock.GetAsync(_processUsagesLock))
```
jaredpar: This part makes me nervous.  The code formatter is optimized for two specific scenarios:

1. For conforming code do as little as possible.
2. For non-conforming code clean up the code as fast as possible

The first issue has higher priority than the second.  The logic being we want the tool to be fast enough that it could be run essentially on build.  

This particular check though does seem fairly expensive as you noted.  I'm worried about the perf impact it's going to have on clean runs of the tool. 

Other rules like the private field naming have the same problem.  It's mitigated by delaying the expensive checks by first doing a lot of inexpensive ones.  I'm trying to brainstorm ways that the same could be done here. 

Going to run this on Roslyn later on to get a sense of the true impact.  Maybe it's not that bad and I'm being over paranoid.

ChadNedzlek: I can add a small optimization if the ConcurrentBag has a size of zero; since that would mean the code was conforming.  Then we don't actually have to rewrite anything, so at least all of the third walk could be skipped.  I'm not sure how to avoid the first two walks to actually tell if the code is conforming, though. The second could be skipped if there were no writable fields anywhere in the input, but that seems highly unlikely.

ChadNedzlek:
I think I've addressed everything I can address.

jaredpar:
@ChadNedzlek sorry, I've gotten caught up in a series of fire drills and not been able to look at this. 

The solution I was mulling over was adding this under a flag like /aggressive, /extra, etc ...  This was a flag we had considered adding already for other reasons:

- The goal of the code formatter is provably safe rewrites.  But some enforcing some of our coding guidelines can't be done in a safe way in a reasonable amount of time.  
- Some rewrites are valuable but inherently expensive.  We want to preserve the speed of the tool for the majority case but allow for the expensive but correct rewrites.  

This particular feature seems to fit that mold to me at least.

ChadNedzlek:
Hrm. Interesting. I could probably take a swing at trying to augment the system to have a sort of "level" of formatting (in a completely different pull request... I don't think I'd want to combine both things into one request).

I'm curious about a few things before I begin.  When would something run the "lite" version of the formatter, and when would someone run the "heavy" version of it?

And is it really a binary choice, or should there be levels of it (things that run in <1sec, <10sec, <1min, <10min, etc)? I can implement it either way, I'm just not 100% of the use cases right now, so I might not have all the context.

jaredpar:
@ChadNedzlek the idea I had in my head, which may not be the same idea everyone else had, is the code formatter could be used in the following cadence:

- Normal use is safe + fast enough to be a completely automated process.  Fast enough to potentially even be used on build or on the same level as running unit tests.  
- Heavy use is done on a cadence basis say once a week, month.  Catches issues that had slipped in during the intervening time.  Hopefully done as a part of an automated job.

ChadNedzlek:
Alrighty. I'm working on adding the option now. I've also created issue #111 to track that.

jaredpar:
@ChadNedzlek thanks for putting up with my crazy.  :smile:

ChadNedzlek:
Now using the /aggressive switch for this formatting rule.

jaredpar:
Looks great!

PR #112, opened 2015-04-15
Add a switch to opt-out of slow checks

### LINKED ISSUE #111: Provide levels of formatting (/simple v /aggressive)
Some of the checks currently implemented and new checks that we want to implement are rather expensive, especially on larger projects.  This means that it becomes awkward to run the formatter in an automated fashion to keep the code base clean on every commit.

If we add an "/agressive" switch, then the fast checks can be run essentially on every build, to keep the code base automatically pristine, while the heavier checks could be run on a slow cadence (or even on demand).

ChadNedzlek:
Some checks that we will soon implement are rather slow
and would prevent using the tool as a normal part of commits.
Create a switch so that slower tests can be opt-in, so that the
default (or simple) experience is fast and simple, so it can be run
frequently.

Fix #111


ChadNedzlek:
@dotnet-bot test this please

ChadNedzlek:
Jeremy pointed out I messed up the indentation of the usage text. Re-pushing.

ChadNedzlek:
@jaredpar Here's the pull request that allows segmenting rules off into "simple" and "agressive" ones.  I don't think we have anything that we would classify as "agressive" yet, but the readonly rule would be the first one.

I used the attribute because this seemed the same sort of information as the "order" piece, and that was how we were encoding that information.

PR #114, opened 2015-04-16
Verify Assert Argument Order

### LINKED ISSUE #105: AssertXxx(actual, expected) where the "actual" is a constant should be reversed
In many tests, Asserts similar to AssertEqual(calcuatedValue, ExpectedConstant) are used.  However, sometimes those are inverted (where the caculatedValue that's being tested is passed to the "expected" value), which leads to strange failure messages that hinder the ability to quickly diagnose failures.

It should be fairly easy to identify the worst offenders, where the "actual" is a constant, and the "expected" is some value calculated in the test itself.

ChadNedzlek:
Test assert methods assume that the "actual" value is a variable
and the "expected" value is a constant for the most part. Inverting
these arguments causes misleading test failure messages that hinder
the ability to quickly diagnose test failures.

Detecting obvious mistakes (where "actual" is constant, and
"expected" is variable) can be done and the arguments swapped.

Fix #105

### DIFF
--- a/src/Microsoft.DotNet.CodeFormatting/Rules/RuleOrder.cs
+++ b/src/Microsoft.DotNet.CodeFormatting/Rules/RuleOrder.cs
@@ -27,6 +27,7 @@ namespace Microsoft.DotNet.CodeFormatting.Rules
         public const int ExplicitVisibilityRule = 2;
         public const int IsFormattedFormattingRule = 3;
         public const int RemoveExplicitThisRule = 4;
+        public const int AssertArgumentOrderRule = 5;
     }
 
     // Please keep these values sorted by number, not rule name.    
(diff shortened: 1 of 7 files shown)


ChadNedzlek:
Anyone have any thoughts about this one?

jaredpar:
This seems like it should be in the XUnit formatter vs. the code formatter.  The code formatter is generally domain agnostic and focussed on coding style.  This is focusing on the correctness of a single domain.

ChadNedzlek:
Where is the "XUnit formatter"? I see the XUnitConverter, but that appears to be specifically about *moving* tests from one framework to another... not something you'd run on existing XUnit code... Though I could be mistaken.

jaredpar:
That is what I meant.  That should be idempotent as well.  

This may also be better as an analyzer / code fix.

ChadNedzlek:
Ah. I didn't realize it was intended to be run on already XUnit projects.  I should be able to move it fairly quickly.

ChadNedzlek:
There. I've moved it to the XUnit converter project.

PR #127, opened 2015-05-29
Add code formatting rule to add newlines at the end of code files

dsplaisted:
@stephentoub has [pointed out](https://github.com/dotnet/corefx/pull/1851#discussion_r31315254) that text files should end with a newline.  This PR adds a code formatting rule to do this for C# source code files.

### DIFF
--- a/src/Microsoft.DotNet.CodeFormatting.Tests/Microsoft.DotNet.CodeFormatting.Tests.csproj
+++ b/src/Microsoft.DotNet.CodeFormatting.Tests/Microsoft.DotNet.CodeFormatting.Tests.csproj
@@ -112,6 +112,7 @@
     <Compile Include="Rules\ExplicitVisibilityRuleTests.cs" />
     <Compile Include="Rules\HasNoIllegalHeadersFormattingRuleTests.cs" />
     <Compile Include="Rules\NewLineAboveRuleTests.cs" />
+    <Compile Include="Rules\NewLineAtEndOfFileRuleTests.cs" />
     <Compile Include="Rules\PrivateFieldNamingRuleTests.cs" />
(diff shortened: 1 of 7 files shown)


khellang:
I guess I'll close #61 then. I never got to finishing it anyways. This closes #54.

dsplaisted:
Thanks for the pointer to #61 @khellang.  Based on reading the discussion there, I've fixed the rule for cases where there are comments or compiler directives at the end of the file.

For simplicity, this PR merely makes sure that the file ends with a newline (in contrast with #61).  It doesn't try to prevent it from ending with multiple newlines.

Priya91:
Ahh i just implemented this before seeing this PR :) 

@dsplaisted: Can you rebase these commits, it's difficult to code review with context in different commits.

jaredpar:
Won't this conflict with the standard formatter rule?  The formatter by default will delete new lines at the end of the file.

jasonwilliams200OK:
Note that this is only required in case of CRLF on Windows. If you are using LF in editor, extra linefeed will persist in diff, which means git doesn't require additional line-break.

I think git converts `CRLF+CRLF` as `LF` for end of file and considers single CRLF as invalid end of file terminator.
One standard way to solve this problem is to have LF rule for all sources in .gitattributes file like this:

```
[*] 
end_of_line = lf

// or

[*.cs, *.vb, *.csproj] 
end_of_line = lf
```

(in corefx repository, [.gitattributes is present](https://github.com/dotnet/corefx/blob/master/.gitattributes), but almost never used: all rules are commented out)

Note, git only apply the .gitattributes rules on `commit`, meaning your local formatting will stay intact.

PR #128, opened 2015-06-04
Update to Roslyn RC

jaredpar:
This updates the solution to use the RC version of Roslyn.

Additionally it removes three uses of ?. (the conditional access
operator).  That was preventing the solution from being compiled in VS
2013.


bartonjs:
The project requirements are a little foofy on the VS compatibility (one could mistakenly read it as VS2015 is required).  If VS2013 is the target (and/or msbuild 12 actually works) then you might want to tighten up the wording on the prereqs.

The change semantically seems fine (aka "LGTM")

jaredpar:
@bartonjs will clarify :) 

The intent is to keep this project usable in VS 2013.  I'm more than happy with moving to VS 2015 once it becomes RTM.  But till then I'd like to keep it usable in the latest RTM version of VS.  Makes it easier for people to contribute :)

PR #136, opened 2015-06-19
build.cmd should always check for NuGet package updates

lgolding:
build.cmd was skipping the NuGet restore command if the
src\packages directory existed at all, regardless of what
was in it. As a result, the command line build would fail
whenever we updated a package (as, for example, I did
yesterday in updating to Roslyn 1.0.0-rc3).

Modify build.cmd to restore packages unconditionally. The
fraction of a second it takes to discover that all packages
are up to date is not significant.

@srivatsn @jaredpar


srivatsn:
LGTM

Priya91:
:+1:

PR #140, opened 2015-06-24
Prepare for analyzer/fixer based formatting operations

lgolding:
The purpose of this change is to open up a spot in the code into which we can fit formatting operations based on Roslyn analyzers/code fixers. The Roslyn-based operations will exist side by side with the existing "rule"-based operations. By default, codeformatter will execute the rule-based operations. In the next commit, we will introduce a command-line switch to perform the analyzer/fixer-based operations instead. When all operations have been re-implemented as analyzer/fixers, we will remove the rule-based operations.

The switching points are in the methods `FormatProjectAsync` and `FormatSolutionAsync` in the `FormattingEngineImplementation` class.

For rule-based and analyzer/fixer-based operations to exist side by side, we must modify codeformatter to use MEF 2 instead of MEF 1, and that is actually the bulk of the changes in this PR. This is necessary because Roslyn code fixers use MEF 2 (because they must be portable, and MEF 1 is not available in portable .NET; only MEF 2 is available). I could have chosen to have codeformatter continue to use MEF 1 for the rule-based operations, and use MEF 2 only for the analyzer/fixer-based operations, but I felt this was too confusing.

Also:
- Remove unused usings and sort usings.
- Fix a bug in build.cmd where it would not update to the latest NuGet packages.

*Note* You might find it helpful to read the individual commit messages for more details on the rationale for some of the changes. The last commit -- a change to the `Options`  class -- is particularly subtle.

*Note* Although I have opened up seams in both the product code and the unit test code to accommodate analyzer/fixer-based operations, this PR does not actually include any such. That will come next.

@Priya91 @srivatsn

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -32,9 +32,7 @@ IF NOT EXIST %LocalAppData%\NuGet md %LocalAppData%\NuGet
 @powershell -NoProfile -ExecutionPolicy unrestricted -Command "$ProgressPreference = 'SilentlyContinue'; Invoke-WebRequest 'https://www.nuget.org/nuget.exe' -OutFile '%CACHED_NUGET%'"
 
 :restore
-IF EXIST "%~dp0src\packages" goto build
+IF NOT EXIST src\packages md src\packages
 %CACHED_NUGET% restore %SOLUTION_PATH%
-
-:build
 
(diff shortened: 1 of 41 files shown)


srivatsn:
LGTM

Priya91:
Rebase away the merge commits (there are 2 merge commits in the list).. Added a few comments, otherwise LGTM.

lgolding:
Thanks Lakshmi. I'll make the changes, rebuild, and merge.

PR #144, opened 2015-06-30
Disable the using location rule by default

jaredpar:
Moving a using from inside a namespace to outside it is a fundamentally dangerous
operation.  In part because it changes the context of the lookup:

- Outside the namespace only the other usings are considered
- Inside the namespace the namespace + the usings are considered

This makes it trivial to create an error by moving a using directive around.

Instead of continuing to patch this rule I've decided to just disable it by
default for now.  We can revisit and make it safer later on but for now it
takes too much work to fix this.


Priya91:
LGTM

PR #145, opened 2015-07-10
Get first analyzer/fixer-based formatter working

lgolding:
This commit includes several distinct changes:

1) Introduce a command line option /useAnalyzers

By default, CodeFormatter continues to use the rule-based formatters. When the
/useAnalyzers command line switch is specified, CodeFormatter uses the new
analyzer/fixer-based formatters instead.

2) Introduce the first analyzer/fixer-based formatter.

This is Sri's "explicit this" analyzer/fixer. Also introduce Sri's
"UberCodeFixer" class, which looks up the fixer associated with each
diagnostic, and registers the fixes.

3) Refactor unit test base classes to account for differences between rules
   and analyzer/fixers

The Format method of CodeFormatterTestBase is now abstract, because it is
implemented differently in rule-base tests and analyzer/fixer-based tests.
The existing rule-based implementation is pushed down into RuleTestBase.
The abstract method RewriteDocumentAsync is also pushed down into RuleTestBase
(where it will still be overridden by the various rule-based test classes),
because analyzer-based tests won't use this method -- because it is only
called from Format, and analyzer/fixer-based tests will Format differently.

TODO: Move RuleTestBase and its three derived classes out of CodeFormatter-
TestBase.cs, into a new file RuleTestBase.cs.

4) Enable test data sharing between rule-based and analyzer/fixer-based tests

The unit tests for the fixers will be identical to the existing tests for the
rules (except that they will derive from a new class AnalyzerFixerTestBase
instead of RuleTestBase), and the two sets of tests will coexist for now.

We avoid duplicating the test data between these two sets of tests by pulling
it out into a set of .NET resource streams in the test assembly. The test data
file follow a naming convention that relates the name of the test classes and
test methods to the file system locations of the data files.

This allows us to add a new overload of CodeFormatterTestBase.Verify,
which allows each test method to be written as a single line, for example:

    [Fact]
    public void TestFieldUse()
    {
        Verify(nameof(TestFieldUse), runFormatter: false);
    }

There is a tradeoff here. Separating out the test data -- in particular,
separating the input data from the expected output data -- means that you have
to look at two files to see what a test actually does. In return, you get
these advantages:

    - It's easy to share test data between rule-based tests and analyzer/fixer
   -based tests.

    - You get Intellisense and syntax highlighting as you write the test data
   files.

    - You don't have to count lines to compute the location of a fix -- just
   place your cursor on it and look at the status bar.

(Source of this idea: Pluralsight couse "Diagnostic Analyzers in Visual Studio
2015: First Look" by Kathleen Dollard,
http://www.pluralsight.com/courses/vs-2015-diagnostic-analyzers-first-look)

5) Move to MEF 2

The Roslyn code fixers are marked with MEF 2 attributes, because they must be
portable, and MEF 1 is not portable. Therefore we change all other uses of MEF
to be MEF 2 as well.

This change broke a few unit tests. These tests (in CombinationTest.cs) first
created a FormattingEngineObject. Since that class's importing constructor
takes an Options object, MEF creates one and passes it to the constructor.
These tests then modified the Options object, for example, by setting its
CopyrightHeader property.

Next, these tests would tell the FormattingEngineObject to perform a format.
This involved instantiating all the formatting rule objects. Each formatting
rule class has an importing constructor that takes an Options object.

Now in MEF 1, where the default part creation policy was "shared", each rule
class was instantiated with the existing Options object. But in MEF 2, where
the default policy is non-shared, each rule was instantiated with a *new*
Options object -- and *not* the one whose properties had been set up for
purposes of the test.

The fix was to mark the Options class with the [Shared] attribute.


lgolding:
@srivatsn @Priya91

srivatsn:
:+1:

Priya91:
Apart from the embedding test cases as separate CS files, everything else LGTM.

jaredpar:
@Priya91 suggestion of constant strings makes sense to me as well.  Everything else LGTM

Priya91:
:+1:

PR #146, opened 2015-07-14
Implement MarkReadonlyFieldsRule as analyzer/fixer + minor cleanups

lgolding:
Use LocalizableString for diagnostic title and message format strings.

Also, various cleanups:
- Pull RuleTestBase into a separate file, per feedback from a previous PR
- Make an internal variable private
- Use simplified property definition syntax


lgolding:
@jaredpar @Priya91 @srivatsn

src/Microsoft.DotNet.CodeFormatting/Rules/UsingLocationRule.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.DotNet.CodeFormatting.Rules
     /// <summary>
     /// This will ensure that using directives are placed outside of the namespace.
     /// </summary>
-    [SyntaxRule(Name = UsingLocationRule.Name, Description = UsingLocationRule.Description, Order = SyntaxRuleOrder.UsingLocationFormattingRule)]
+    [SyntaxRule(Name = UsingLocationRule.Name, Description = UsingLocationRule.Description, Order = SyntaxRuleOrder.UsingLocationFormattingRule, DefaultRule = false)]
```
lgolding: When I was cherry-picking the four commits in this PR from another local branch onto my local anfix branch, this line came up as a conflict, and I chose the version with DefaultRule=false. It looks like that was wrong and that I had no reason to change this file.

lgolding:
Added the MarkReadonlyFieldsRule -- which, when broken into an analyzer and a fixer, I think is better named UnwrittenWritableFieldsAnalyzer and UnwrittenWritableFieldsFixer.

src/CodeFormatter/CommandLineParser.cs
```diff
@@ -32,6 +32,18 @@ public sealed class CommandLineOptions
             verbose: false,
             useAnalyzers: false);
 
+        public static readonly CommandLineOptions ListAnalyzers = new CommandLineOptions(
```
lgolding: A new "canned" set of command line options to support the newly added /analyzers option. This lists the available analyzers, by analogy to /rules, which lists the available rules. Actually what it really lists is the available diagnostics, so maybe it should be "/diagnostics" instead. Let me know if you prefer that.

src/Microsoft.DotNet.CodeFormatting.Tests/AnalyzerFixerTestBase.cs
```diff
@@ -12,10 +12,31 @@ public abstract class AnalyzerFixerTestBase : CodeFormattingTestBase
     { 
         private IFormattingEngine _engine; 
 
-        protected AnalyzerFixerTestBase(IFormattingEngine engine)
```
lgolding: Previously, I had each derived class create an engine and pass it down to the base class through this ctor. But the derived classes don't need the engine, so let the base class take care of it. This simplifies the derived classes.

src/Microsoft.DotNet.CodeFormatting.Tests/AnalyzerFixerTestBase.cs
```diff
@@ -12,10 +12,31 @@ public abstract class AnalyzerFixerTestBase : CodeFormattingTestBase
...
+                if (_engine == null)
+                {
+                    _engine = FormattingEngine.Create();
+                }
+
+                return _engine;
+            }
+        }
+
+        protected void DisableAllDiagnostics()
```
lgolding: Methods to enable derived classes to enable only the one diagnostic they are testing. The rule-based test classes didn't need this mechanism because they worked by explicitly calling ProcessAsync on the single rule they were testing. But the analyzer-based tests work by calling (ultimately) CompilationWithAnalyzers.GetAnalyzerDiagnosticsAsync -- so I have to make sure that only the desired analyzer gets into the CompilationWithAnalyzers.

src/Microsoft.DotNet.CodeFormatting.Tests/CodeFormattingTestBase.cs
```diff
@@ -97,73 +93,4 @@ protected void Verify(string source, string expected, bool runFormatter = true,
         }
 
     }
-
-    // TODO: Pull RuleTestBase and its three derived classes into separate file RuleTestBase.cs.
```
lgolding: The rest of this file is pulled into a separate file RuleTestBase.cs. No other code change (except to remove now-unused usings).

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitThisAnalyzer.cs
```diff
@@ -15,17 +16,16 @@ namespace Microsoft.DotNet.CodeFormatting.Analyzers
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
     public class ExplicitThisAnalyzer : DiagnosticAnalyzer
     {
-        internal static DiagnosticDescriptor rule = new DiagnosticDescriptor("DNS0001",
```
lgolding: Make the string available to the fixer so it doesn't have to repeat the hard-coded strings.

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitThisAnalyzer.cs
```diff
@@ -15,17 +16,16 @@ namespace Microsoft.DotNet.CodeFormatting.Analyzers
...
-                                                                             "Don't use explicit 'this' for private fields",
-                                                                             "Don't use explicit 'this' for private fields",
```
lgolding: Move to resources

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitThisAnalyzer.cs
```diff
@@ -39,7 +39,7 @@ public override void Initialize(AnalysisContext context)
                         node.Expression.Kind() == SyntaxKind.ThisExpression &&
                         IsPrivateField(node, syntaxContext.SemanticModel, syntaxContext.CancellationToken))
                     {
-                        syntaxContext.ReportDiagnostic(Diagnostic.Create(rule, node.GetLocation()));
+                        syntaxContext.ReportDiagnostic(Diagnostic.Create(rule, node.GetLocation(), node.Name));
```
lgolding: Make diagnostic more informative.

src/Microsoft.DotNet.CodeFormatting/FormattingEngine.cs
```diff
@@ -24,6 +26,13 @@ public static ImmutableArray<IRuleMetadata> GetFormattingRules()
             return engine.AllRules;
         }
 
+        public static ImmutableArray<DiagnosticDescriptor> GetSupportedDiagnostics()
```
lgolding: This method supports the /analyzers command line option.

src/Microsoft.DotNet.CodeFormatting/FormattingEngineImplementation.cs
```diff
@@ -33,8 +33,9 @@ internal sealed partial class FormattingEngineImplementation : IFormattingEngine
         private readonly IEnumerable<ExportFactory<ILocalSemanticFormattingRule, LocalSemanticRule>> _localSemanticRules;
         private readonly IEnumerable<ExportFactory<IGlobalSemanticFormattingRule, GlobalSemanticRule>> _globalSemanticRules;
         private readonly Stopwatch _watch = new Stopwatch();
-        private readonly Dictionary<string, bool> _ruleMap = new Dictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
-        private readonly ImmutableDictionary<string, CodeFixProvider> _fixerMap;
+        private readonly Dictionary<string, bool> _ruleEnabledMap = new Dictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
+        private readonly Dictionary<string, bool> _diagnosticEnabledMap;
+        private readonly ImmutableDictionary<string, CodeFixProvider> _diagnosticIdToFixerMap;
```
lgolding: I did some renaming here. In Sri's code he had a "fixerMap" but it served a different purpose than the rule map. I renamed all the maps to clarify their purpose.

src/Microsoft.DotNet.CodeFormatting.Tests/RuleTestBase.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Formatting;
+using Microsoft.DotNet.CodeFormatting;
+using Microsoft.DotNet.CodeFormatting.Tests;
+
+public abstract class RuleTestBase : CodeFormattingTestBase
```
jaredpar: Should have a namespace in this file

src/Microsoft.DotNet.CodeFormatting/Analyzers/UnwrittenWritableFieldAnalyzer.cs
```diff
@@ -0,0 +1,351 @@
...
+    {
+        internal const string DiagnosticId = "DNS0002";
+        private static DiagnosticDescriptor rule = new DiagnosticDescriptor(DiagnosticId,
+                                                                            ResourceHelper.MakeLocalizableString(nameof(Resources.UnwrittenWritableFieldAnalyzer_Title)),
+                                                                            ResourceHelper.MakeLocalizableString(nameof(Resources.UnwrittenWritableFieldAnalyzer_MessageFormat)),
+                                                                            "Usage",
+                                                                            DiagnosticSeverity.Warning,
+                                                                            true);
+
+        private static readonly SyntaxKind[] compoundAssignmentExpressionKinds =
```
jaredpar: Private static field should be prefixed with `s_`

src/Microsoft.DotNet.CodeFormatting/Analyzers/UnwrittenWritableFieldAnalyzer.cs
```diff
@@ -0,0 +1,351 @@
...
+                }
+
+                node = node.Parent;
+            }
+
+            return false;
+        }
+    }
+
+    internal static class FieldSymbolExtensions
```
jaredpar: Would separate out these types to different files if they are generally useful types.

src/Microsoft.DotNet.CodeFormatting/Analyzers/UnwrittenWritableFieldFixer.cs
```diff
@@ -0,0 +1,59 @@
...
+                CodeAction.Create(
+                    Resources.UnwrittenWritableFieldFixer_Title,
+                    c => AddReadonlyModifier(context.Document, root, fieldDeclarationNode)),
+                diagnostic);
+        }
+
+        private Task<Document> AddReadonlyModifier(Document document, SyntaxNode root, FieldDeclarationSyntax fieldDeclaration)
+        {
+            FieldDeclarationSyntax newFieldDeclaration = fieldDeclaration
+                .WithModifiers(fieldDeclaration.Modifiers.Add(s_readOnlyToken));
```
jaredpar: Do we have a test to see what happens if the field is prefixed with a comment?  Consider:

``` csharp
/// <summary> stuff </summary>
int m_field;
```

In the past the `WithModifiers` method has gotten me in trouble because it would put the modifier before the comment instead of after which is the more logical place.  I believe I got into the most troubleu here with methods though so the same problem may not apply to fields.  And maybe it is limited to non-XML comments but either way just want to check if we have a test here.

src/Microsoft.DotNet.CodeFormatting.Tests/RuleTestBase.cs
lgolding: Fixed

src/Microsoft.DotNet.CodeFormatting/Analyzers/UnwrittenWritableFieldAnalyzer.cs
lgolding: Fixed.

jaredpar:
Overall LGTM.  Added some minor feedback

src/Microsoft.DotNet.CodeFormatting/Analyzers/UnwrittenWritableFieldAnalyzer.cs
lgolding: I don't know yet that they are generally useful. These were originally private methods within the rule class, like `IsFieldSerializableByAttribute(IFieldSymbol)`, which I re-cast as extension methods because I like being able to write `if (symbol.IsSerializableByAttributes())`. If they come in useful elsewhere I'll extract them.

src/Microsoft.DotNet.CodeFormatting/Analyzers/UnwrittenWritableFieldFixer.cs
lgolding: Added tests for both XML and non-XML comments:
````csharp
class C
{
    /// <summary>Stuff</summary>
    private READONLY int read;
}
````
and
````csharp
class C
{
    // Stuff
    private READONLY int read;
}
````
 Both passed.

jaredpar: Awesome.  Thanks for adding those tests.

lgolding: Not so awesome :) See above. I edited my comment in place when I discovered that it fails if you omit the access specifier.

jaredpar: Ah yes that was the case that bit me before.

PR #149, opened 2015-07-21
Upgrade to Roslyn 1.0.0

genlu:
- Add Microsoft.CodeAnalysis.CSharp.Analyzers and Microsoft.CodeAnalysis.Analyzers to Microsoft.DotNet.CodeFormatting project, which was installed but never used.

@jaredpar


jaredpar:
LGTM

PR #154, opened 2015-07-28
Copy CopyrightHeader.md and IllegalHeaders.md to the correct output location

### LINKED ISSUE #103: CopyrightHeader.md is not copied to the right bin location during a build
Using a fresh enlistment in CodeFormatter (or git clean out the bin folders) run build from the root.
Now execute CodeFormatter.exe against some .SLN file.
Expected: runs correctly
Actual: errors reported about missing CopyrightHeaders.md

It looks like CopyrightHeaders is being written to bin\codeformatter\debug only, but CodeFormatter.exe is written to bin.  IllegalHeaders.md is written to the correct location.

jaredmoo:
Changed CopyrightHeader.md and IllegalHeaders.md from <None> items to <Content> items because they are files that are used by the output binaries. Use msbuild's builtin <CopyToOutputDirectory> instead of having a manual copy step. Fixes #103.

### DIFF
--- a/src/CodeFormatter/CodeFormatter.csproj
+++ b/src/CodeFormatter/CodeFormatter.csproj
@@ -83,9 +83,15 @@
     <Compile Include="Properties\AssemblyInfo.cs" />
   </ItemGroup>
   <ItemGroup>
+    <Content Include="CopyrightHeader.md">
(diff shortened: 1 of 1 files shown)


jaredmoo:
Can this be merged?

PR #155, opened 2015-07-28
Add ExplicitVariableTypeAnalyzer and ExplicitVariableTypeFixer

genlu:
- `ExplicitVariableTypeAnalyzer` reports a diagnostic when keyword 'var' is used in variable declaration where it could be replaced by an explicit type.
- `ExplicitVariableTypeFixer` replaces the usage of 'var' reported by `ExplicitVariableTypeAnalyzer` with an explicit type.
- `ExplicitVariableTypeAnalyzer `won't fire a warning when it is obvious what the type is (this is what in .NET C# coding style). 
    - Here's the complete list of such scenarios (declared type is "obvious"):
        - `LiteralExpressionSyntax`: `var foo = 10;`
        - `CastExpressionSyntax`: `var x = (Foo)foo;`
        - `ObjectCreationExpressionSyntax`: `var foo = new Foo();`
        - `ArrayCreationExpressionSyntax`: `var x = new bool[] { true, false };`
        - `BinaryExpressionSyntax` and `Kind` is `SyntaxKind.AsExpression`: `var x = foo as Bar;`
    - Here's an incomplete list of scenarios the analyzer won't fire (declared type is considered non-obvious):
        - `var x = Foo();`
        - `var y = new[] { 1, 2, 3 }; //implicit array creation`
        - `var z = foo;  // foo is another variable`

@lgolding @jaredpar @Priya91 @michaelcfanning @srivatsn


src/Microsoft.DotNet.CodeFormatting.Tests/Analyzers/ExplicitVariableTypeAnalyzerTests.cs
```diff
@@ -0,0 +1,237 @@
...
+
+namespace Microsoft.DotNet.CodeFormatting.Tests
+{
+    public sealed class ExplicitVariableTypeAnalyzerTests : AnalyzerFixerTestBase
+    {
+        public ExplicitVariableTypeAnalyzerTests()
+        {
+            DisableAllDiagnostics();
+            EnableDiagnostic(ExplicitVariableTypeAnalyzer.DiagnosticId);
+        }
```
lgolding: Blank line after closing brace of method.

src/Microsoft.DotNet.CodeFormatting.Tests/Analyzers/ExplicitVariableTypeAnalyzerTests.cs
```diff
@@ -0,0 +1,237 @@
...
+class C1
+{
+    void M()
+    {
+        int x = 0;
+        bool[] y = { true };  
+        int[][] z = new[] { new[] { 1, 2, 3 }, new[] { 4, 5, 6 } };
+    }
+}";
+            var expected = @"
```
lgolding: Since you expect no change, you can save typing by just setting expected = test. I had to read the whole "expected" string to convince myself there was no change.

Also I'd make those strings const. (I see I didn't do that in mine, either.)

src/Microsoft.DotNet.CodeFormatting.Tests/Analyzers/ExplicitVariableTypeAnalyzerTests.cs
```diff
@@ -0,0 +1,237 @@
...
+class C1
+{
+    void M()
+    {
+        Int32 x = 0;
+        Boolean[] y = new[] { true, false };  
+        Int32[][] z = new[] { new[] { 1, 2, 3 }, new[] { 4, 5, 6 } };
+        var anon = new { Name = ""Terry"", Age = 34 };
```
lgolding: I think for clarity this last one is a separate test: TestAnonymousTypeDeclaration.

src/Microsoft.DotNet.CodeFormatting.Tests/Analyzers/ExplicitVariableTypeAnalyzerTests.cs
```diff
@@ -0,0 +1,237 @@
...
+    void M()
+    {
+        U1 x = new U1();
+        U2 y = new U2();
+        U3 z = U3.E1;
+        using (U1 x = new U1())
+        { }         
+    }
+}";
+            var expected = @"
```
lgolding: Again, expected = text would be clearer, or even omit the variable and say

````
Verify(text, text, runFormatter: false);
````

File: src/Microsoft.DotNet.CodeFormatting.Tests/Analyzers/ExplicitVariableTypeAnalyzerTests.cs [hunk 5]
```diff
@@ -0,0 +1,237 @@
...
+    {}
+
+    enum U3 { E1, E2 }
+
+    void M()
+    {
+        var x = new U1();
+        var y = new U2();
+        var z = U3.E1;
+        using (var x = new U1())
```
lgolding: Again I think this is clearer as a separate test: TestVarDeclarationInUsing

src/Microsoft.DotNet.CodeFormatting.Tests/Analyzers/ExplicitVariableTypeAnalyzerTests.cs
```diff
@@ -0,0 +1,237 @@
...
+ 
+        foreach (string[] element in a2)
+        { }
+    }
+}";
+            Verify(text, expected, runFormatter: false);
+        }
+
+        [Fact]
+        public void TestSimpleForEachVarDeclaration()
```
lgolding: Are there other places you could test, like

````
for(var i = 0; i < max; ++i)
````

File: src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeAnalyzer.cs [hunk 7]
```diff
@@ -0,0 +1,81 @@
...
+using System.Linq;
+using System.Collections.Generic;
+
+namespace Microsoft.DotNet.CodeFormatting.Analyzers
+{
+    [Export(typeof(DiagnosticAnalyzer))]
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    class ExplicitVariableTypeAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "DNS0099";
```
lgolding: Any reason not to use DNS0003?

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeAnalyzer.cs
```diff
@@ -0,0 +1,81 @@
...
+        internal const string VariableDeclarationCustomTag = "VariableDeclarationTag";
+        internal const string ForEachStatementCustomTag = "ForEachStatementTag";
+
+        private static DiagnosticDescriptor ruleVariableDeclaration = new DiagnosticDescriptor(DiagnosticId,
```
lgolding: CodeFormatter will want you name static variables with `s_`.

Also, CodeFormatter will notice that these can be made `readonly`.

src/Microsoft.DotNet.CodeFormatting/Resources.resx
```diff
@@ -126,6 +126,15 @@
   <data name="ExplicitThisFixer_Title" xml:space="preserve">
     <value>Remove 'this' qualifier</value>
   </data>
+  <data name="ExplicitVariableTypeAnalyzer_MessageFormat" xml:space="preserve">
+    <value>Use VariableTypeDeclaration</value>
+  </data>
+  <data name="ExplicitVariableTypeAnalyzer_Title" xml:space="preserve">
+    <value>Variable '{0}' should be declared with explicit type </value>
```
lgolding: Remove trailing space after "type".

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeAnalyzer.cs
genlu: Just thought you might have those reserved for existing rules. Should I take 0003?

src/Microsoft.DotNet.CodeFormatting.Tests/Analyzers/ExplicitVariableTypeAnalyzerTests.cs
genlu: @lgolding I think it is just a `VariableDeclarationSyntax` in a for loop, but sure, I will add one for it.

src/Microsoft.DotNet.CodeFormatting.Tests/Analyzers/ExplicitVariableTypeAnalyzerTests.cs
genlu: @lgolding Sure. Will fix. (and other identical strings)

src/Microsoft.DotNet.CodeFormatting.Tests/Analyzers/ExplicitVariableTypeAnalyzerTests.cs
genlu: @lgolding OK.

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeAnalyzer.cs
```diff
@@ -0,0 +1,81 @@
...
+        private static DiagnosticDescriptor ruleForEachStatement = new DiagnosticDescriptor(DiagnosticId,
+                                                                            ResourceHelper.MakeLocalizableString(nameof(Resources.ExplicitVariableTypeAnalyzer_Title)),
+                                                                            ResourceHelper.MakeLocalizableString(nameof(Resources.ExplicitVariableTypeAnalyzer_MessageFormat)),
+                                                                            "Style",
+                                                                            DiagnosticSeverity.Warning,
+                                                                            true,
+                                                                            customTags: ForEachStatementCustomTag);
+
+
+        private static readonly ImmutableArray<DiagnosticDescriptor> s_supportedRules = ImmutableArray.Create(ruleVariableDeclaration, ruleForEachStatement);
```
lgolding: Blank line below.

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeAnalyzer.cs
```diff
@@ -0,0 +1,81 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
+        {
+            get
```
lgolding: Consider using C# 6.0 expression body syntax:

````
... SupportedDiagnostics
    => s_supportedRules;
````

File: src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeAnalyzer.cs [hunk 12]
```diff
@@ -0,0 +1,81 @@
...
+                return s_supportedRules;
+            }
+        }
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterSyntaxNodeAction(syntaxContext =>
+            {
+                var node = (VariableDeclarationSyntax)syntaxContext.Node; 
+                if (node.Type.IsVar && !IsAnonymousType(node.Variables.Single(), syntaxContext.SemanticModel, syntaxContext.CancellationToken))
```
lgolding: Add comment to explain that this is ok because `var` can't have multiple variable dedclarators.

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeFixer.cs
```diff
@@ -0,0 +1,153 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Simplification;
+using Microsoft.CodeAnalysis.Text;
+using Microsoft.CodeAnalysis.CSharp;
+using System.Threading;
```
lgolding: Put with the other `System.` directives.

src/Microsoft.DotNet.CodeFormatting.Tests/Analyzers/ExplicitVariableTypeAnalyzerTests.cs
genlu: @lgolding Fixed.

src/Microsoft.DotNet.CodeFormatting.Tests/Analyzers/ExplicitVariableTypeAnalyzerTests.cs
genlu: @lgolding Fixed.

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeAnalyzer.cs
genlu: @lgolding Taking DNS0003 as discussed offline.

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeFixer.cs
```diff
@@ -0,0 +1,153 @@
...
+
+namespace Microsoft.DotNet.CodeFormatting.Analyzers
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
+    class ExplicitVariableTypeFixer : CodeFixProvider
```
lgolding: `public`

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeAnalyzer.cs
genlu: @lgolding Fixed.

src/Microsoft.DotNet.CodeFormatting/Resources.resx
genlu: @lgolding Fixed.

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeAnalyzer.cs
genlu: @lgolding Fixed.

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeAnalyzer.cs
genlu: @lgolding  OK.

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeFixer.cs
```diff
@@ -0,0 +1,153 @@
...
+            {
+                context.RegisterCodeFix(
+                    CodeAction.Create(
+                        Resources.ExplicitVariableTypeFixer_Title,
+                        c => ReplaceVarWithExplicitType(context.Document, root, oldTypeSyntaxNode, newTypeSyntaxNode.WithTriviaFrom(oldTypeSyntaxNode))),
+                    diagnostic);
+            }
+        }
+
+        private Task<Document> ReplaceVarWithExplicitType(Document document, SyntaxNode root, SyntaxNode varNode, SyntaxNode ExplicitTypeNode)
```
lgolding: `ExplicitTypeNode` variable name should be camelCase `explicitTypeNode`.

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeFixer.cs
```diff
@@ -0,0 +1,153 @@
...
+        {
+            return Task.FromResult(
+                document.WithSyntaxRoot(root.ReplaceNode(varNode, ExplicitTypeNode.WithAdditionalAnnotations(Simplifier.Annotation))));
+        }
+
+        // return null if could not construct a TypeSyntax with explicit type.
+        private static TypeSyntax GetTypeSyntaxNode(SyntaxNode node, SemanticModel model, CancellationToken cancellationToken)
```
lgolding: I'd call this `Make...` or `Create...` because to me "`Get`" implies that the thing already exists.

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeFixer.cs
```diff
@@ -0,0 +1,153 @@
...
+                return null;
+            }
+
+            var symbol = (ILocalSymbol)model.GetDeclaredSymbol(node, cancellationToken); 
+            ITypeSymbol typeSymbol = symbol.Type;
+            // typeSymbol.IsAnonymousType is guaranteed to be false since we already filtered out anonymous type in analyzer
+            Debug.Assert(!typeSymbol.IsAnonymousType);
+            TypeSyntax newTypeSyntaxNode = null;
+            switch (typeSymbol.TypeKind)
+            {
```
lgolding: @jaredpar is there a higher-level way to do this? This seems like very low-level syntax manipulation.

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeFixer.cs
genlu: @lgolding Fixed.

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeFixer.cs
genlu: @lgolding Fixed.

lgolding:
LGTM, modulo my picky comments, and the one substantive question for @jaredpar.

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeFixer.cs
genlu: @lgolding Ah, yes.

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeFixer.cs
genlu: @lgolding Sounds good. Changed to `CreateTypeSyntaxNode`.

src/Microsoft.DotNet.CodeFormatting/Analyzers/ExplicitVariableTypeFixer.cs
```diff
@@ -0,0 +1,154 @@
...
+                        Resources.ExplicitVariableTypeFixer_Title,
+                        c => ReplaceVarWithExplicitType(context.Document, root, oldTypeSyntaxNode, newTypeSyntaxNode.WithTriviaFrom(oldTypeSyntaxNode))),
+                    diagnostic);
+            }
+        }
+
+        private Task<Document> ReplaceVarWithExplicitType(Document document, SyntaxNode root, SyntaxNode varNode, SyntaxNode explicitTypeNode)
+        {
+            return Task.FromResult(
+                document.WithSyntaxRoot(root.ReplaceNode(varNode, explicitTypeNode.WithAdditionalAnnotations(Simplifier.Annotation))));
```
genlu: @jaredpar @Priya91 So this doesn't simplify system types like `Int32` and `String` to keyword `int` and `string`. Did I do anything wrong here?

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeFixer.cs
```diff
@@ -0,0 +1,114 @@
...
+                    typeSyntaxNode = forEachStatementNode.Type;
+                    varSymbol = model.GetDeclaredSymbol(forEachStatementNode, cancellationToken);
+                    break;
+
+                case RuleType.None:
+                    break;
+            }
+            if (varSymbol != null)
+            {
+                DocumentEditor documentEditor = await DocumentEditor.CreateAsync(context.Document, cancellationToken);context.RegisterCodeFix(
```
srivatsn: You can move the creation of the DocumentEditor into the ReplaceVarWithExplicittype method and call TypeExpression(..) in there. It's generally good practice to defer most of the work to inside the delegate for the codeaction because that only runs when the actual fix is invoked as opposed to any code outside that will have to run when we need to determine if there is any fix at all.

srivatsn: missing newline for the context.Register.. statement

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeFixer.cs
```diff
@@ -0,0 +1,114 @@
...
+                                                        root, 
+                                                        typeSyntaxNode, 
+                                                        documentEditor.Generator.TypeExpression(varSymbol.Type).WithTriviaFrom(typeSyntaxNode))),
+                    diagnostic);
+            }
+        }
+
+        private Task<Document> ReplaceVarWithExplicitType(Document document, SyntaxNode root, SyntaxNode varNode, SyntaxNode explicitTypeNode)
+        {
+            return Task.FromResult(
```
srivatsn: You could simply call documentEdtior.ReplaceNode and then return documentEdtior.GetchangedDocument()

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeFixer.cs
```diff
@@ -0,0 +1,114 @@
...
+            switch (GetRuleType(diagnostic))
+            {
+                case RuleType.RuleVariableDeclaration:
+                    var variableDeclarationNode = root.
+                        FindNode(diagnosticSpan).
+                        FirstAncestorOrSelf<VariableDeclarationSyntax>();
+                    Debug.Assert(variableDeclarationNode != null);
+                    typeSyntaxNode = variableDeclarationNode.Type;
+                    // Implicit typed variables cannot have multiple declartors
+                    varSymbol = (ILocalSymbol)model.GetDeclaredSymbol(variableDeclarationNode.Variables.Single(), cancellationToken);
```
srivatsn: Do you want to do a check to see if this is implicitly declared? Otherwise .Single() would throw an exception.

genlu: @srivatsn The analyzer only triggers on implicit declarations, is the check still necessary here? i.e. Is it possible for the fixer to see a different AST than the analyzer?

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeFixer.cs
```diff
@@ -0,0 +1,114 @@
...
+                        FirstAncestorOrSelf<ForEachStatementSyntax>();
+                    Debug.Assert(forEachStatementNode != null);
+                    typeSyntaxNode = forEachStatementNode.Type;
+                    varSymbol = model.GetDeclaredSymbol(forEachStatementNode, cancellationToken);
+                    break;
+
+                case RuleType.None:
+                    break;
+            }
+            if (varSymbol != null)
```
srivatsn: check for null for typeSyntaxNode as well. In incomplete code scenarios, you could have that be null for a variable.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeFixer.cs
```diff
@@ -0,0 +1,114 @@
...
+                    Debug.Assert(variableDeclarationNode != null);
+                    typeSyntaxNode = variableDeclarationNode.Type;
+                    // Implicit typed variables cannot have multiple declartors
+                    varSymbol = (ILocalSymbol)model.GetDeclaredSymbol(variableDeclarationNode.Variables.Single(), cancellationToken);
+                    break;
+
+                case RuleType.RuleForEachStatement:
+                    var forEachStatementNode = root.
+                        FindToken(diagnosticSpan.Start).
+                        Parent.
```
srivatsn: Is Parent guaranteed to be non-null? I'd use .Parent?.FirstAncestor....

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        // Return true if given SyntaxNode has an anonymous type, which can't be replaced by an explicit type.
+        private static bool IsAnonymousType(SyntaxNode node, SemanticModel model, CancellationToken cancellationToken)
+        {
+            ISymbol symbol = model.GetDeclaredSymbol(node, cancellationToken); 
+            return ((ILocalSymbol)symbol).Type.IsAnonymousType;
+        }
+
+        // Return true if given SyntaxNode is either VariableDeclarationSyntax or ForEachStatementSyntax and
```
srivatsn: Use XML doc comments instead of regular comments.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+                if (node.Type.IsVar &&
+                    !IsTypeObvious(node) &&
+                    !IsAnonymousType(node, syntaxContext.SemanticModel, syntaxContext.CancellationToken))
+                {
+                    syntaxContext.ReportDiagnostic(Diagnostic.Create(s_ruleForEachStatement, node.Identifier.GetLocation(), node.Identifier.Text));
+                }
+            }, SyntaxKind.ForEachStatement);
+        }
+
+        // Return true if given SyntaxNode has an anonymous type, which can't be replaced by an explicit type.
```
srivatsn: Use XML doc comments.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+        private static bool IsAnonymousType(SyntaxNode node, SemanticModel model, CancellationToken cancellationToken)
+        {
+            ISymbol symbol = model.GetDeclaredSymbol(node, cancellationToken); 
+            return ((ILocalSymbol)symbol).Type.IsAnonymousType;
```
srivatsn: Use  ?.s liberally. In incomplete code situations, any of this could be null.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeFixer.cs
srivatsn: Ah ok .that's fine then.

src/Microsoft.DotNet.CodeFormatter.Analyzers.Tests/ExplicitVariableTypeAnalyzerTests.cs
```diff
@@ -0,0 +1,408 @@
...
+            const string expected = @"
+class C1
+{
+    void M()
+    {
+        var x = 0;
+        var y = new bool[] { true, false };
+    }
+}";
+            Verify(text, expected, runFormatter: false);
```
lgolding: test = expected in this test.

src/Microsoft.DotNet.CodeFormatter.Analyzers.Tests/ExplicitVariableTypeAnalyzerTests.cs
```diff
@@ -0,0 +1,408 @@
...
+        {
+            const string text = @"
+class C1
+{
+    void M()
+    {
+        foreach (var element in new string[] { "" })
+        { }  
+ 
+        foreach (var element in new int[] { new int[] { 1, 2, 3 }, new int[] { 4, 5, 6 } })
```
lgolding: In test `TestSimpleNonobviousTypeVarDeclaration` above, you had a similar `var` declaration initialized with a nested initializer, and you replaced `var` with `int[][]`. Why in the `foreach` case is it ok for it to remain `var`?

genlu: @lgolding In the test you mentioned, it is `var z = new[] { new[] { 1, 2, 3 }, new[] { 4, 5, 6 } };`, which is an anonymous array creation such that the type is not given directly in the declaration like a regular array declaration (even though it is kinda obvious in this case).

lgolding: @genlu Ah, right.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.DotNet.CodeFormatter.Analyzers
+{
+    [Export(typeof(DiagnosticAnalyzer))]
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    class ExplicitVariableTypeAnalyzer : DiagnosticAnalyzer
```
lgolding: Do analyzers need to be marked `public`? (If not, explicitly mark it `internal`.)

genlu: @lgolding Ah.. I missed this one. Thanks for notice that. Since @michaelcfanning has refactored the code such that analyzers are now in a separate binary, I think it should be `public`.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        // Return true if given SyntaxNode has an anonymous type, which can't be replaced by an explicit type.
+        private static bool IsAnonymousType(SyntaxNode node, SemanticModel model, CancellationToken cancellationToken)
+        {
+            ISymbol symbol = model.GetDeclaredSymbol(node, cancellationToken); 
+            return ((ILocalSymbol)symbol).Type.IsAnonymousType;
+        }
+
+        // Return true if given SyntaxNode is either VariableDeclarationSyntax or ForEachStatementSyntax and 
+        // VariableDeclarationSyntax.Variables.Single().Initializer.Value or ForEachStatementSyntax.Expression is:
```
lgolding: This puts a lot of implementation into the comment. I'd replace the text on this line (that is, the text following the "and" on the previous line) with:

"... the variable is initialized with one of the following:"

src/Microsoft.DotNet.CodeFormatter.Analyzers/Resources.resx
```diff
@@ -126,6 +126,15 @@
...
     <value>Remove 'this' qualifier</value>
   </data>
+  <data name="ExplicitVariableTypeAnalyzer_MessageFormat" xml:space="preserve">
+    <value>Variable '{0}' should be declared with explicit type</value>
+  </data>
+  <data name="ExplicitVariableTypeAnalyzer_Title" xml:space="preserve">
+    <value>Replace the usage of 'var' with an explicit type</value>
+  </data>
+  <data name="ExplicitVariableTypeFixer_Title" xml:space="preserve">
+    <value>Replace the usage of 'var' with explicit type</value>
```
lgolding: "... with *an* explicit type" (for consistency with the fixer title). And I think the message format should also be ".. with *an* explicit type".

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+        //          - ArrayCreationExpressionSyntax
+        // 
+        //      ImplicitArrayCreationExpressionSyntax: This one is not included. e.g. new[] {}
+        //
+        // TODO: AnonymousObjectCreationExpressionSyntax could be filtered out here as well, maybe we want to do that?
+        //       The trade-off here is the logic we use to check syntax node is not as accurate as a query to SemanticModel, 
+        //       which is (presumbly) much slower
+        private static bool IsTypeObvious(SyntaxNode node)
+        {
+
```
lgolding: Remove extra blank line.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+            
+            if (node == null)
+            {
+                return false;
+            }
+
+            ExpressionSyntax expressionNode = null;
+            if (node is VariableDeclarationSyntax)
+            {
+                expressionNode = ((VariableDeclarationSyntax)node).Variables.Single().Initializer.Value;
```
lgolding: FxCop would complain about the inefficiency of casting twice (once in the `is` expression, once with the explicit cast. Instead:
````
var variableDeclaration = node as VariableDeclarationSyntax;
if (variableDeclaration != null)
{
    expressionNode = variableDeclaration.Variables.Single().Initializer.Value;
    ...
````
````

File: src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeFixer.cs [hunk 34]
```diff
@@ -0,0 +1,114 @@
...
+
+            foreach (string customTag in diagnostic.Descriptor.CustomTags)
+            {
+                switch (customTag)
+                {
+                    case ExplicitVariableTypeAnalyzer.VariableDeclarationCustomTag:
+                        return RuleType.RuleVariableDeclaration;
+                    case ExplicitVariableTypeAnalyzer.ForEachStatementCustomTag:
+                        return RuleType.RuleForEachStatement;
+                    // Diagnostics corresponding to this fixer must has either one of these above two custom tags
```
lgolding: "has" => "have"

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeFixer.cs
```diff
@@ -0,0 +1,114 @@
...
+            ILocalSymbol varSymbol = null;
+            switch (GetRuleType(diagnostic))
+            {
+                case RuleType.RuleVariableDeclaration:
+                    var variableDeclarationNode = root.
+                        FindNode(diagnosticSpan).
+                        FirstAncestorOrSelf<VariableDeclarationSyntax>();
+                    Debug.Assert(variableDeclarationNode != null);
+                    typeSyntaxNode = variableDeclarationNode.Type;
+                    // Implicit typed variables cannot have multiple declartors
```
lgolding: "declarators"

lgolding:
LGTM with only minor quibbles (and the `public` declaration, which you have already addressed).

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
genlu: @lgolding Although it does 2 extra casting, I think the if - elseif makes the logic clearer.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeFixer.cs
genlu: Fixed.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeFixer.cs
genlu: @srivatsn Hmm... in this case, I think the null check should be in analyzer, and fixer wouldn't need to check again.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeFixer.cs
genlu: @srivatsn The span provided by diagnostic is retrieved by `ForEachStatementSyntax.Identifier.GetLocation()`, and the existence of the ForEachStatementSyntax node guarantees the Parent is non-null I think. On the other hand, I probably need to do a null check for Identifier property in the analyzer...

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
genlu: Fixed.

src/Microsoft.DotNet.CodeFormatter.Analyzers/Resources.resx
genlu: Fixed.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
genlu: Fixed.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeFixer.cs
genlu: Fixed.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeFixer.cs
genlu: Fixed.

src/Microsoft.DotNet.CodeFormatter.Analyzers.Tests/ExplicitVariableTypeAnalyzerTests.cs
genlu: Fixed.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
genlu: Fixed.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
genlu: Fixed.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeFixer.cs
genlu: Fixed.

genlu:
@michaelcfanning @srivatsn Ping...

src/Microsoft.DotNet.CodeFormatter.Analyzers.Tests/ExplicitVariableTypeAnalyzerTests.cs
```diff
@@ -0,0 +1,399 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.DotNet.CodeFormatter.Analyzers;
+
+using Xunit;
+
+namespace Microsoft.DotNet.CodeFormatter.Analyzers.Tests
+{
+    public sealed class ExplicitVariableTypeAnalyzerTests : AnalyzerFixerTestBase
```
michaelcfanning: I think something like 'ProvideExplicitVariableType' might be a more complete/descriptive name

src/Microsoft.DotNet.CodeFormatter.Analyzers.Tests/ExplicitVariableTypeAnalyzerTests.cs
```diff
@@ -0,0 +1,399 @@
...
+    {
+        public ExplicitVariableTypeAnalyzerTests()
+        {
+            DisableAllDiagnostics();
```
michaelcfanning: very likely disabling all diagnostics is something common to these tests that could be in a test base class. something to think about for a future change

src/Microsoft.DotNet.CodeFormatter.Analyzers.Tests/ExplicitVariableTypeAnalyzerTests.cs
```diff
@@ -0,0 +1,399 @@
...
+        int[][] z = new[] { new[] { 1, 2, 3 }, new[] { 4, 5, 6 } };
+    }
+}"; 
+            Verify(text, text, runFormatter: false);
+        }
+
+        [Fact]
+        public void TestSimpleObviousTypeVarDeclaration()
+        {
+            const string text = @"
```
michaelcfanning: consider using 'input' as a name instead of text. also suggest always creating an 'expected' string variable. set expected = input; where that makes sense. alternately you could name your parameters in your verify call, like so:

Verify(input, expected : input, runFormatter: false)

I think the former would be better. this is a nit. your tests are very nice

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+    public class ExplicitVariableTypeAnalyzer : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = AnalyzerIds.ExplicitVariableType;
+        internal const string VariableDeclarationCustomTag = "VariableDeclarationTag";
+        internal const string ForEachStatementCustomTag = "ForEachStatementTag";
+
+        private static readonly DiagnosticDescriptor s_ruleVariableDeclaration = new DiagnosticDescriptor(DiagnosticId,
+                                                                            ResourceHelper.MakeLocalizableString(nameof(Resources.ExplicitVariableTypeAnalyzer_Title)),
+                                                                            ResourceHelper.MakeLocalizableString(nameof(Resources.ExplicitVariableTypeAnalyzer_MessageFormat)),
+                                                                            "Style",
```
michaelcfanning: this needs to be put into a const string class at some point.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+        private static readonly ImmutableArray<DiagnosticDescriptor> s_supportedRules = ImmutableArray.Create(s_ruleVariableDeclaration, s_ruleForEachStatement);
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
+            => s_supportedRules;
+                                              
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterSyntaxNodeAction(syntaxContext =>
+            {
+                var node = (VariableDeclarationSyntax)syntaxContext.Node;
+                // C# syntax doesn't allow implicitly typed variables have multiple declartors,
```
michaelcfanning: misspelling here and s/be 'to have multiple declarators'?

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+            {
+                return false;
+            }
+
+            ExpressionSyntax expressionNode = null;
+            if (node is VariableDeclarationSyntax)
+            {
+                expressionNode = ((VariableDeclarationSyntax)node)?.Variables.FirstOrDefault()?.Initializer?.Value;
+            }
+            else if (node is ForEachStatementSyntax)
```
michaelcfanning: I would just use else here. the cast to ForEachStatementSyntax would throw an exception in the following clause if it is the wrong type (because you know it is non-null). So the Debug.Fail would be unnecessary

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
```diff
@@ -0,0 +1,134 @@
...
+            {
+                Debug.Fail("This program location is thought to be unreachable.");
+            }
+            // 'expressionNode == null' means the code under inspection has errors,
+            // so we return 'true' to avoid firing a warning.
+            return expressionNode == null ?
+                   true :
+                   expressionNode != null &&
+                 ((expressionNode is BinaryExpressionSyntax &&
+                   expressionNode.Kind() == SyntaxKind.AsExpression) ||
```
michaelcfanning: don't you just need the SyntaxKind.AsExpression test here? IE, is it ever possible that Syntax.AsExpression is true but the item is not a BinaryExpressionSyntax? Probably not, I'm guessing

src/Microsoft.DotNet.CodeFormatter.Analyzers.Tests/ExplicitVariableTypeAnalyzerTests.cs
genlu: @michaelcfanning Sounds good. Fixed.

src/Microsoft.DotNet.CodeFormatter.Analyzers.Tests/ExplicitVariableTypeAnalyzerTests.cs
genlu: @michaelcfanning Fixed.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
genlu: @michaelcfanning You are right. Fixed.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ExplicitVariableTypeAnalyzer.cs
genlu: @michaelcfanning Fixed.

srivatsn:
:+1:

PR #157, opened 2015-07-29
Anfix refactor analyzers

michaelcfanning:
This change breaks the analyzers and associated tests into a separate assembly. The formatting engine helpers were updated to pass a set of assemblies for use in the composition. In the next change, per discussion with JaredPar, I will update the tool to use a standard command-parsing library. After this, we'll update the command-line to accept an arbitrary set of analyzers to apply to the formatting operation.

The preview below also includes the 'remove unused imports' analyzer, which is submitted as a PR. If that gets approved soon, I won't have to fix my mistake and delete this mistakenly included content. :)


michaelcfanning:
@lgolding @jaredpar @Priya91

jaredpar:
:+1:

srivatsn:
:+1:

src/Microsoft.DotNet.CodeFormatter.Analyzers/NameHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// </summary>
+    internal static class NameHelper
+    {
+        /// <summary>
+        /// Get the full name of a symbol. So a field might look like
+        /// "OuterNamespace.Inner.ClassName.FieldName"
+        /// </summary>
+        /// <param name="symbol">symbol to get name of</param>
+        /// <returns>Full display name for a symbol</returns>
+        internal static string GetFullName(ISymbol symbol)
```
lgolding: Sri says there is a class `SymbolDisplay` that has a method to do this.

michaelcfanning: ok. This file not changed during this PR only moved. let's look at this in future

lgolding:
:+1:

src/Microsoft.DotNet.CodeFormatter.Analyzers/NameHelper.cs
srivatsn: Ya one can use `symbol.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat)` to get the full name. I'll merge this is in now since you just moved the file.

PR #158, opened 2015-07-29
Carry System.Threading.Tasks.Dataflow.dll with tool, as this dependen…

michaelcfanning:
…cy is not installed with .NET 4.6


michaelcfanning:
@jaredpar @Priya91

michaelcfanning:
@lgolding

jaredpar:
:+1: 

Didn't realize we had that dependency

src/CodeFormatter/CodeFormatter.csproj
```diff
@@ -71,6 +71,10 @@
       <HintPath>..\packages\System.Reflection.Metadata.1.0.21\lib\portable-net45+win8\System.Reflection.Metadata.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="System.Threading.Tasks.Dataflow, Version=4.5.24.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\References\System.Threading.Tasks.Dataflow.dll</HintPath>
```
akoeplinger: Is there a reason you're checking in the binary as opposed to using the NuGet package?

michaelcfanning: This issue is apparently related to a bug in MSBUILD where this indirect dependency isn't properly copied during build to the output directory. This change is a temporary fix until that problem is resolved (which will be relatively soon). So I opted for this expedient change (which I will revert when the msbuild problem is resolved).

akoeplinger: That shouldn't matter, you can just add System.Threading.Tasks.Dataflow to packages.config and reference it like S.R.Metadata above. This way it's no longer an indirect dependency as well.

michaelcfanning: Right, and I wouldn't be required to make another change in future. Will do this.

srivatsn: +1 to using the package.

lgolding: +2. Checked-in binaries are particularly bad with Git because when you clone a repo you get its entire history. So you get all the versions of all the binaries that have ever been checked in, even if they're not used any more.

michaelcfanning: Yes. I haven't succeeded in getting the recent update of this dependency to download from nuget. I will wait and sort this out, though.

https://www.nuget.org/packages/System.Threading.Tasks.Dataflow/

PR #159, opened 2015-07-29
Add Gen's 'prefer explicit type to var' analyzer and fixer.

michaelcfanning:
@lgolding @genlu @srivatsn @Priya91 @jaredpar


jaredpar:
This rule goes against the guidance used by many of the teams that use this formatter.  We prefer the opposite (`var` over explicit type in many cases).

michaelcfanning:
sounds like a good candidate rule for parameterization. with codeformatter preferring aggressive use of var in default use. i'll follow up offline to get a sense of the standard you enforce.

genlu:
@jaredpar @michaelcfanning  Just add some clarifications, this analyzer/fixer does some filtering on the implicit declaration, and "only" (probably still more than people would like it to) fire on cases where the type is not quite obvious. Here's some example.

- Won't fire:
    - `var i = 99;`
    - `var arr = new string[][]`
    - `var foo = new Foo();`
    - `var bar = (Bar)foo;
    - `var baz = bar as Baz;`

- Will fire (examples from scanning codeformatter itself):
    - `var container = CreateCompositionContainer(assemblies);`
    - `foreach (var fixer in _fixers){...}`
    - `var document = this.CurrentSolution.GetDocument(documentId);`
    - `var firstUsing = syntaxRoot.DescendantNodesAndSelf().OfType<UsingDirectiveSyntax>().FirstOrDefault();`

As you can see, the filter can be tweaked to filter out more cases if needed, which can be configured by user.

@michaelcfanning   I have updated another PR I opened which covers all the changes in this one, and has more contexts on what's going on. I guess you can just close this one.

The existing PR is:
https://github.com/dotnet/codeformatter/pull/155

jaredpar:
@genlu gotcha.  Even with those exceptions though it would still cause a substantial amount of Roslyn code to change that we wouldn't like.  It's probably something that would need to be off in that Repo.

genlu:
@jaredpar Agree. This analyzer definitely needs to be configured on a per project basis. I think @lgolding and @michaelcfanning are implementing a configuration mechanism that can turn this analyzer on/off on demand/by default.

michaelcfanning:
10.We only use  var  when it's obvious what the variable type is (i.e.  var stream = new FileStream(...)  not  var stream = OpenStandardInput() ).

@jaredpar : the text above is from your coding standards document, which from limited description seems to match what gen proposes to implement. I will take gen's analyzer and run it on Roslyn to see what it uncovers, I'm curious.

jaredpar:
@michaelcfanning it's probably the most contentious rule in the document.  :frowning:  At some point though we do need to rationalize it.

PR #160, opened 2015-07-30
Add explicit reference to System.Threading.Tasks.Dataflow package, v4…

michaelcfanning:
Adds an explicit reference to the System.Threading.Tasks.Dataflow package, v4.5.25, recently released. Works around a current MSBUILD issue where this indirect dependency doesn't get copied to the build output directory (with the result that xcopy deployments can fail due to missing assembly).


michaelcfanning:
@jaredpar @Priya91

michaelcfanning:
@srivatsn

srivatsn:
:+1:

PR #161, opened 2015-07-31
Anfix commandline parser

michaelcfanning:
Changes:

-- Picked up dependency on pre-release CommandLine library (which provides better generics support and other benefits)
-- Split program function into 'list' and 'format' commands'
-- 'list' reports metadata on built-in rules and analyzers
-- 'format' invokes formatting against one or more format targets

Breaks from previous version:
-- Library doesn't support multiple instances of a single argument.
-- No forward slash support for args
-- All targets must be specified with an explicit -t or --targets command
-- rule+- syntax replaced with /enable and /disable commands that accept one or more rule names (comma-delimited)
-- /tables command renamed to --definedotnetformatter

Open questions:
  1) what about --define-dotnet_formatter as an argument name?
  2) verify that library doesn't support multiple instances of same arg

Once I get some preliminary sign-off, I will update docs and solicit review. I will include a section on migrating older command-lines to new version, with examples


michaelcfanning:
@jaredpar @Priya91 @srivatsn @lgolding @genlu

src/CodeFormatter/FormatOptions.cs
```diff
@@ -0,0 +1,99 @@
...
+        public IEnumerable<string> PreprocessorConfigurations { get; set; }
+
+        [Option(
+            "copyright", 
+            HelpText = "Specifies file containing copyright header.")]
+        public string CopyrightHeader { get; set;  }
+
+        [Option(
+            "enable", 
+            HelpText = "Comma-separated list of rules to disable",
```
lgolding: enable

src/CodeFormatter/FormatOptions.cs
```diff
@@ -0,0 +1,99 @@
...
+        public IEnumerable<string> Files { get; set; }
+
+        [Option(
+            'l', "lang", 
+            HelpText = "Specifies the language to use when a response file is specified, e.g., 'C#', 'Visual Basic', ... (default: 'C#').")]
+        public string Language { get; set; }
+
+        [Option(
+            'c', "config", 
+            HelpText = "Additional preprocessor configurations the formatter should run under.",
```
lgolding: "Comma-separated list of additional..."

src/Microsoft.DotNet.CodeFormatting/IFormattingEngine.cs
```diff
@@ -16,7 +16,7 @@ public interface IFormattingEngine
         ImmutableArray<string> FileNames { get; set; }
         ImmutableArray<IRuleMetadata> AllRules { get; }
         ImmutableArray<DiagnosticDescriptor> AllSupportedDiagnostics { get; }
-        bool AllowTables { get; set; }
+        bool AllowTables { set; }
```
lgolding: A write-only property?

src/CodeFormatter/Program.cs
```diff
@@ -18,46 +20,43 @@ namespace CodeFormatter
...
                                         new Assembly[] {
                                             typeof(FormattingEngine).Assembly,
                                             typeof(OptimizeNamespaceImportsAnalyzer).Assembly
                                         };
 
         private static int Main(string[] args)
+        {            
+            return Parser.Default.ParseArguments<ListOptions, FormatOptions>(args)
+              .Return(
+                (ListOptions listOptions) => RunListCommandAndReturnExitCode(listOptions),
```
lgolding: Is the reason for including "AndReturnExitCode" in the method names to alert the reader that CommandLine will call `exit` with the return code from this method? I'm not usually a fan of "And" in method names because it suggests the method is doing more than one thing and should be refactored, but I guess it's ok here.

src/CodeFormatter/Program.cs
```diff
@@ -18,46 +20,43 @@ namespace CodeFormatter
...
-                    RunListRules(options.UseAnalyzers);
-                    exitCode = 0;
-                    break;
-                case Operation.Format:
-                    exitCode = RunFormat(options);
-                    break;
-                default:
-                    throw new Exception("Invalid enum value: " + options.Operation);
-            }
+            ListRulesAndAnalyzer(options.Analyzers, options.Rules);
```
lgolding: `ListRulesAndAnalyzers`

src/Microsoft.DotNet.CodeFormatting.Tests/Microsoft.DotNet.CodeFormatting.Tests.csproj
```diff
@@ -84,7 +88,7 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="CodeFormattingTestBase.cs" />
-    <Compile Include="CommandLineParserTests.cs" />
+    <Compile Include="FormatOptionsParsingTests.cs" />
```
lgolding: I don't see `FormatOptionsParsingTests.cs` in this pull request.

src/CodeFormatter/FormatOptions.cs
michaelcfanning: fixed thx

src/CodeFormatter/FormatOptions.cs
michaelcfanning: yes, fixed, thx

src/Microsoft.DotNet.CodeFormatting/IFormattingEngine.cs
michaelcfanning: oops. was looking at dependencies on the write. this property probably warrants a rename but I want to verify with @jared or @Priya91 that I get what it's used for.

src/CodeFormatter/Program.cs
michaelcfanning: pattern taken from commandline samples. but I agree with you, changed

src/CodeFormatter/Program.cs
michaelcfanning: fixed

src/Microsoft.DotNet.CodeFormatting.Tests/Microsoft.DotNet.CodeFormatting.Tests.csproj
michaelcfanning: file was untracked, thanks, resolved

src/Microsoft.DotNet.CodeFormatting.Tests/FormatOptionsParsingTests.cs
```diff
@@ -0,0 +1,152 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Collections.Generic;
+
+using CodeFormatter;
```
lgolding: Do we have a standard for spelling out the namespaces?

src/Microsoft.DotNet.CodeFormatting/IFormattingEngine.cs
```diff
@@ -16,7 +16,7 @@ public interface IFormattingEngine
         ImmutableArray<string> FileNames { get; set; }
         ImmutableArray<IRuleMetadata> AllRules { get; }
         ImmutableArray<DiagnosticDescriptor> AllSupportedDiagnostics { get; }
-        bool AllowTables { get; set; }
+        bool AllowTables { get;  set; }
```
lgolding: Only change to this file is now introduction of extraneous white space. Just revert the file.

src/Microsoft.DotNet.CodeFormatting.Tests/FormatOptionsParsingTests.cs
```diff
@@ -0,0 +1,152 @@
...
+namespace Microsoft.DotNet.CodeFormatting.Tests
+{
+    public class FormatOptionsParsingTests
+    {
+        private FormatOptions Parse(out int result, params string[] args)
+        {
+            result = 1;
+            FormatOptions options = null;
+
+            using (var writer = new StringWriter())
```
lgolding: It doesn't look like `writer` is used anywhere.

src/Microsoft.DotNet.CodeFormatting.Tests/FormatOptionsParsingTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            int result;
+
+            // CodeFormatter format --enable enabledRule -t test.csproj
+            var options = Parse(out result, "format", "--enable", "enabledRule", "-t", "test.csproj");
+
+            Assert.Equal(0, result);
+            Assert.NotNull(options);
+            Assert.True(options.RuleMap["enabledRule"]);
+            Assert.True(options.RuleMap.Count == 1);
```
lgolding: I think equalities are more clearly asserted with `Assert.Equal(1, options.RuleMap.Count)` (here and below)

src/Microsoft.DotNet.CodeFormatting.Tests/FormatOptionsParsingTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        public void MultipleTargets()
+        {
+            int result;
+
+            // CodeFormatter format --enable enabledRule
+            var options = Parse(out result, "format", "-t", "projectOne.csproj", "projectTwo.csproj");
+
+            Assert.Equal(0, result);
+            Assert.NotNull(options);
+            Assert.Equal(2, options.FormatTargets.Count());
```
lgolding: Add test of contents? `Assert.True(options.FormatTargets.Contains(projectOne.csproj)`, etc.

src/Microsoft.DotNet.CodeFormatting.Tests/FormatOptionsParsingTests.cs
michaelcfanning: would be glad to make a scrub for this later but let's loop in jared/Priya. this code is a rename/updating of existing tests, so i'll leave the NS alone for now

src/Microsoft.DotNet.CodeFormatting.Tests/FormatOptionsParsingTests.cs
michaelcfanning: yes, good catch, vestiges of a broken attempt to fix a test multi-threading issue

src/Microsoft.DotNet.CodeFormatting.Tests/FormatOptionsParsingTests.cs
michaelcfanning: ok

src/Microsoft.DotNet.CodeFormatting.Tests/FormatOptionsParsingTests.cs
michaelcfanning: yessir

lgolding:
:+1:

PR #163, opened 2015-08-02
Fix UnwrittenWritableFieldAnalyzer with tests

genlu:
The bug is caused by not identifying increment/decrement operators as write operation.

@lgolding @srivatsn


src/Microsoft.DotNet.CodeFormatter.Analyzers/UnwrittenWritableFieldAnalyzer.cs
```diff
@@ -44,6 +44,14 @@ public class UnwrittenWritableFieldAnalyzer : DiagnosticAnalyzer
                 SyntaxKind.SubtractAssignmentExpression
             };
 
+        private static readonly SyntaxKind[] s_unaryAssignmentKinds =
```
AlexGhiondea: Could you have merged these with the compound operators?

genlu: @AlexGhiondea Sure.

AlexGhiondea:
LGTM

srivatsn:
:+1:

PR #164, opened 2015-08-03
Anfix fix copyright default

michaelcfanning:
Command-line parsing library update broke the copyright header option. This is due to the fact that we don't have unit-testing that covers the driver -> code formatter rules and analyzers path. I will fix this after discussing approach with the team. In the meantime, here's a fix for the copyright header issue.

@lgolding @srivatsn @Priya91 @jaredpar


srivatsn:
:+1:

PR #166, opened 2015-08-07
Fix ProvideExplicitVariableTypeAnalyzer

genlu:
Fix `ProvideExplicitVariableTypeAnalyzer` to handle errors properly
- Make analyzer ignore the node with miss nodes in sub-tree.
- Add corresponding tests.

@tmeschter @srivatsn @michaelcfanning @lgolding @jaredpar @Priya91

### DIFF
--- a/src/Microsoft.DotNet.CodeFormatter.Analyzers/ProvideExplicitVariableTypeAnalyzer.cs
+++ b/src/Microsoft.DotNet.CodeFormatter.Analyzers/ProvideExplicitVariableTypeAnalyzer.cs
@@ -101,9 +101,13 @@ namespace Microsoft.DotNet.CodeFormatter.Analyzers
         //       which is (presumbly) much slower. 
         private static bool IsTypeObvious(SyntaxNode node)
         {            
-            if (node == null)
+            // Since parser would generate the missing node (IsMissing == true) and attach a diagnostic to it,
+            // we traverse the sub tree only if a diagnostic is attached as optimization
+            if (node == null || 
+               (node.ContainsDiagnostics && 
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.CodeFormatter.Analyzers/ProvideExplicitVariableTypeAnalyzer.cs
```diff
@@ -101,9 +101,13 @@ private static bool IsAnonymousType(SyntaxNode node, SemanticModel model, Cancel
         //       which is (presumbly) much slower. 
         private static bool IsTypeObvious(SyntaxNode node)
         {            
-            if (node == null)
+            // Since parser would generate the missing node (IsMissing == true) and attach a diagnostic to it,
+            // we traverse the sub tree only if a diagnostic is attached as optimization
+            if (node == null || 
+               (node.ContainsDiagnostics && 
+                node.DescendantNodesAndTokensAndSelf().Where(n => n.IsMissing).Any()))
```
michaelcfanning: I am sure we want to minimize node descendants traversal. I suggest you rewrite this code a bit to defer this call until later in the method, when you've conclusively determined you would otherwise return false. Make sense? IOW, eliminate these checks here. Evaluate the rest of the function. If retval appears to be false, put in this logic to sanity check you don't need to flip the result to true due to the missing node case you've identified.

genlu: @michaelcfanning Fixed. Thanks!

srivatsn:
:+1:

genlu:
@dotnet-bot Retest this, please.

PR #167, opened 2015-08-09
Add check for invalid file.

paulvanbrenk:
Instead of throwing an exception later, check if we can access the file specified as proj/solution/etc.


jaredpar:
This change doesn't prevent the later exception.  It is always possible for the file to be delete after this check.  In general `File.Exists` or `FileInfo.Exists` is an API that should be avoided at all costs (it really should be deleted from the framework).  

I think we should focus on the place where we actually read the file contents.  If the message / behavior is undesirable there then we should focus on improving it.

NickCraver:
+1 on @jaredpar's comment - that's the appropriate place to handle these things and eliminates the races.

PR #169, opened 2015-08-18
Preserve leading trivia when deleting TestClassAttribute

rainersigwald:
I noticed diffs of the form:

```
-    /// <summary>
-    /// Verify the functioning of the BuildErrorEventArg class.
-    /// </summary>
-    [TestClass]
```

Where not only was the not-needed-in-xunit attribute removed, but its
removal also deleted class-level doc comments.  This change preserves
leading trivia on the deleted AttributeList.  That trivia is normally
blank (so this doesn't break the existing scenario), but a leading comment
or doc comment will be present there.


michaelcfanning:
LGTM

genlu:
:+1:

PR #178, opened 2015-09-21
Skip projects that don't have any documents.

lgolding:
This prevents a "sequence has no elements" exception when we access `projects.Documents.First()`.


lgolding:
@jinujoseph @natidea FYI

PR #180, opened 2015-09-23
Drive all enabled/disabled settings from XML-based configuration files

michaelcfanning:
This change results in all enabled/disabled state, for both analyzers and formatting rules, being driven from options XML. There are two checked in files now (for test) AllEnabled.formatconfig and AllDisabled.format config, that provide examples options files. These files can be generated from the command-line by using the options command, e.g., CodeFormatter.exe options MySettings.formatconfig. Note that the analyzers currently won't pick up options XML files unless they have the .formatconfig suffix. After approving this change, we should change Coby to look for a Coby.formatconfig file that lives alongside Coby.rc. If this file is there, it should be passed to the CodeFormatter.exe format command via the -options-file-path command. Before we get too far with consuming the XML settings format, I will author a JSON serialization/deserialization capability to the property bag library.

@lgolding @jinujoseph @genlu


src/CodeFormatter/FormatOptions.cs
```diff
@@ -50,18 +53,6 @@ internal class FormatOptions
         public string CopyrightHeaderFile { get; set;  }
 
         [Option(
-            "enable",
```
michaelcfanning: These command-line arguments are entirely replaced by XML-driven configuration

src/CodeFormatter/FormatOptions.cs
```diff
@@ -117,21 +108,24 @@ private static ImmutableArray<string> InitializeCopyrightHeaderText(string copyr
...
-                foreach (string rule in EnabledRules)
-                {
-                    _ruleMap = _ruleMap.SetItem(rule, true);
-                }
+                propertyBag.LoadFrom(OptionsFilePath);
             }
 
-            if (DisabledRules != null)
+            propertyBag = (PropertyBag)propertyBag["CodeFormatterRules.Options"];
+            foreach (string key in propertyBag.Keys)
```
michaelcfanning: This kind of direct access of a property bag is not typical (usually, you use a descriptor object as a key). In this case, we're munging the formatting rules code to use this mechanism. This should wither away once all code is converted to analyzers.

src/Microsoft.DotNet.CodeFormatter.Analyzers/Microsoft.DotNet.CodeFormatter.Analyzers.csproj
```diff
@@ -77,7 +77,6 @@
     <Compile Include="OptimizeNamespaceImportsFixer.cs" />
     <Compile Include="OptimizeNamespaceImportsOptions.cs" />
     <Compile Include="OptionsHelper.cs" />
-    <Compile Include="ProvideExplicitVariableTypeOptions.cs" />
```
michaelcfanning: These options files used to contain a single option only, 'Enabled'. Instead of requiring these be explicitly declared/maintained, the tool now generates an Enabled option for every check.

src/CodeFormatter/FormatOptions.cs
```diff
@@ -117,21 +108,24 @@ private static ImmutableArray<string> InitializeCopyrightHeaderText(string copyr
         private ImmutableDictionary<string, bool> BuildRuleMapFromOptions()
         {
             _ruleMap = ImmutableDictionary<string, bool>.Empty;
-            if (EnabledRules != null)
+
```
lgolding: Rather than using an `ImmutableDictionary` (which you then have to recreate each time you call `SetItem` on it, how about using a plain `Dictionary` and then making an `ImmutableDictionary` from it at the end?

michaelcfanning: yes, of course, will do.

src/CodeFormatter/Program.cs
```diff
@@ -46,19 +41,8 @@ private static int Main(string[] args)
...
-            // Currently, we perform discovery against the built-in CodeFormatter rules
-            // and analyzers only.
-            foreach (IOptionsProvider provider in FormattingEngine.GetOptionsProviders(s_defaultCompositionAssemblies))
-            {
-                foreach (IOption option in provider.GetOptions())
-                {
-                    allOptions.SetProperty(option, option.DefaultValue);
-                }
-            }
             allOptions.SaveTo(exportOptions.OutputPath, id: "codeformatter-options");
```
lgolding: I don't understand -- the export options command now emits an empty property bag into the `codeformatter-options` portion of the specified output file?

michaelcfanning: I just moved this logic to a helper, should be doing the same thing.

src/CodeFormatter/Program.cs
```diff
@@ -89,7 +73,7 @@ private static void ListRulesAndAnalyzers(bool listAnalyzers, bool listRules)
 
             if (listAnalyzers)
             {
-                ImmutableArray<DiagnosticDescriptor> diagnosticDescriptors = FormattingEngine.GetSupportedDiagnostics(s_defaultCompositionAssemblies);
+                ImmutableArray<DiagnosticDescriptor> diagnosticDescriptors = FormattingEngine.GetSupportedDiagnostics(OptionsHelper.s_defaultCompositionAssemblies);
```
lgolding: Although private/protected static fields are named with `s_`, I don't think public ones should be. `OptionsHelper.DefaultCompositionAssemblies`?

src/CodeFormatter/Program.cs
lgolding: Oh, I interpreted "Default" to mean "Empty".

src/Microsoft.DotNet.CodeFormatter.Analyzers.Tests/AnalyzerFixerTestBase.cs
```diff
@@ -33,23 +33,21 @@ private IFormattingEngine Engine
             }
         }
 
-        protected void DisableAllDiagnostics()
+        protected PropertyBag BuildAllRulesDisabledPolicy()
```
lgolding: Curious, why did you prefer this name?

genlu: "Rules" makes me think of codeformatter rules, maybe change it to "Analyzers"?

lgolding: "`BuildAllAnalyzersDisabledPolicy`". Bleah ;-) I thought `DisableAllDiagnostics` was already a perfectly cromulent name.

src/CodeFormatter/Program.cs
michaelcfanning: oops, moved this member and neglected a rename. will fix

src/Microsoft.DotNet.CodeFormatter.Analyzers.Tests/AnalyzerFixerTestBase.cs
michaelcfanning: ConstructPolicyThatDisablesAllAnalyzers

Perfectly descriptive and irritatingly prolix. ship it?

genlu:
LGTM

src/Microsoft.DotNet.CodeFormatter.Analyzers.Tests/AnalyzerFixerTestBase.cs
michaelcfanning: Changed this to 'CreatePolicyThatDisablesAllAnalysis'

PR #183, opened 2015-09-30
Regions Suck!

cbjugstad:


jaredpar:
Thanks for sending this contribution.  There are a couple of general feedback points:

- The terminology Suck is not appropriate for this tool.  Please just a more appropriate name. 
- This is not a part of the core .NET coding guidelines and hence needs to be disabled by default.

PR #188, opened 2015-10-16
Upgrade to CommandLineParser 2.0.273-beta

lgolding:
... for consistency with Coby. The reason this matters is that both Coby tools and CodeFormatter.exe are uploaded to Azure for use by Coby; all the binaries go to the same directory, so they need to all use the same versions of packages that they have in common.

@michaelcfanning @jinujoseph

### DIFF
--- a/src/CodeFormatter/CodeFormatter.csproj
+++ b/src/CodeFormatter/CodeFormatter.csproj
@@ -17,8 +17,8 @@
     <CodeAnalysisRuleSet>AllRules.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="CommandLine, Version=2.0.207.0, Culture=neutral, PublicKeyToken=de6f01bd326f8c32, processorArchitecture=MSIL">
-      <HintPath>..\packages\CommandLineParser.2.0.207-alpha\lib\net45\CommandLine.dll</HintPath>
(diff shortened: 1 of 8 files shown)


src/CodeFormatter/CodeFormatter.csproj
```diff
@@ -96,7 +96,9 @@
     <None Include="App.config" />
     <None Include="CopyrightHeader.md" />
     <None Include="IllegalHeaders.md" />
-    <None Include="packages.config" />
+    <None Include="packages.config">
+      <SubType>Designer</SubType>
+    </None>
```
lgolding: Change made automatically by VS. I won't fight it.

src/CodeFormatter/Program.cs
```diff
@@ -26,12 +26,12 @@ internal static class Program
...
         private static int Main(string[] args)
-        {            
+        {
             return Parser.Default.ParseArguments<
-                ListOptions, 
+                ListOptions,
                 ExportOptions,
                 FormatOptions>(args)
-              .Return(
+            .MapResult(
```
lgolding: Breaking API change from alpha to Beta.

michaelcfanning:
LGTM

akoeplinger:
What's Coby?

lgolding:
It's an internal project I'm working on. Sorry, it was indiscreet of me to mention it here :-)

akoeplinger:
Ok thanks :) I've seen it in a few other issues/PRs so I was curious.

PR #193, opened 2015-11-06
Filter rules and analyzer analysis based on project language, where a…

michaelcfanning:
…pplicable.

### DIFF
--- a/src/CodeFormatter/CodeFormatter.csproj
+++ b/src/CodeFormatter/CodeFormatter.csproj
@@ -17,8 +17,8 @@
     <CodeAnalysisRuleSet>AllRules.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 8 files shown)


michaelcfanning:
@lgolding @genlu @srivatsn

src/Microsoft.DotNet.CodeFormatting.Tests/Rules/MarkReadonlyFieldTests.cs
```diff
@@ -384,6 +384,21 @@ public void M(C1 c)
             Verify(Original(text), Readonly(text), true, LanguageNames.CSharp);
         }
 
+        [Fact]
+        public void MarkReadOnlyDoNotAnalyzeVisualBasicCode()
```
srivatsn: Is this rule not applicable for VB? Isn't this test going to break when we implement this rule for VB? A more generic test might be to have a test-only analyzer that is csharp only and test that on VB code.

PR #194, opened 2015-11-06
Added a rule to remove #region/#endregion directives.

jakesays-old:
Rule to remove regions.

### DIFF
--- a/src/Microsoft.DotNet.CodeFormatting/Rules/RemoveRegionsRule.cs
+++ b/src/Microsoft.DotNet.CodeFormatting/Rules/RemoveRegionsRule.cs
@@ -1,0 +1,52 @@
+﻿// Copyright(c) Microsoft.All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Microsoft.DotNet.CodeFormatting.Rules
+{
+	[SyntaxRule(RemoveRegionsRule.Name, RemoveRegionsRule.Description, SyntaxRuleOrder.RemoveRegionsRule)]
+	internal sealed class RemoveRegionsRule : CSharpOnlyFormattingRule, ISyntaxFormattingRule
+	{
+		internal const string Name = "RemoveRegions";
+		internal const string Description = "Removes all regions";
+
+		public SyntaxNode Process(SyntaxNode targetNode, string languageName)
+		{
+			var finder = new RegionFinder();
+			finder.Visit(targetNode);
+			var results = finder.Results;
+
+			if (results.Count > 0)
+			{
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.CodeFormatting/Rules/RemoveRegionsRule.cs
```diff
@@ -0,0 +1,52 @@
...
+
+			if (results.Count > 0)
+			{
+				return targetNode.ReplaceTrivia(results,
+					(arg1, arg2) => new SyntaxTrivia());
+			}
+			return targetNode;
+		}
+
+		class RegionFinder : CSharpSyntaxWalker
```
terrajobst: Should have explicit visibility

terrajobst:
LGTM

@jaredpar?

src/Microsoft.DotNet.CodeFormatting/Rules/RemoveRegionsRule.cs
```diff
@@ -0,0 +1,52 @@
...
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Microsoft.DotNet.CodeFormatting.Rules
+{
+    [SyntaxRule(RemoveRegionsRule.Name, RemoveRegionsRule.Description, SyntaxRuleOrder.RemoveRegionsRule)]
```
jaredpar: I believe this rule should be marked as `DefaultRule=false`.  I agree some people will find value in this rule but it's not a part of our explicit coding guidelines and it would be a great surprise if all regions were suddenly deleted from a lot of projects.

src/Microsoft.DotNet.CodeFormatting/Rules/RemoveRegionsRule.cs
```diff
@@ -0,0 +1,52 @@
...
+            public List<SyntaxTrivia> Results { get; } = new List<SyntaxTrivia>();
+
+            public RegionFinder()
+                : base(SyntaxWalkerDepth.StructuredTrivia)
+            {
+            }
+
+            public override void VisitRegionDirectiveTrivia(RegionDirectiveTriviaSyntax node)
+            {
+                Results.Add(node.ParentTrivia);
```
jaredpar: Why do you track the parent trivia here instead of the node directly?

jaredpar:
I do think this needs to be off by default but other than that it looks good.  

Had 1 Q about the trivia tracking.  More out of curiosity than a correctness issue.

jakesays:
It was the only way I could find to go from the node to a SyntaxTrivia. It seems to work, but if there is a more correct way then I'll gladly change it.

How do I configure the rule to be off by default?

jaredpar:
@jakesays see the end of the attribute declaration here

https://github.com/dotnet/codeformatter/blob/master/src/Microsoft.DotNet.CodeFormatting/Rules/MarkReadonlyFieldsRule.cs#L19

Essentially `DefaultRule = false` is the current method.

PR #197, opened 2015-11-13
Resolve false positives related to failure to recognize pre- and post…

michaelcfanning:
…-increment field operations as mutating relevant members. + tests.

@lgolding @jaredpar @Priya91 @genlu @srivatsn

### DIFF
--- a/src/Microsoft.DotNet.CodeFormatting/Rules/MarkReadonlyFieldsRule.cs
+++ b/src/Microsoft.DotNet.CodeFormatting/Rules/MarkReadonlyFieldsRule.cs
@@ -249,6 +249,20 @@ namespace Microsoft.DotNet.CodeFormatting.Rules
                 CheckForFieldWrite(node.Left);
             }
 
+            public override void VisitPostfixUnaryExpression(PostfixUnaryExpressionSyntax node)
+            {
+                base.VisitPostfixUnaryExpression(node);
+
+                CheckForFieldWrite(node.Operand);
+            }
+
+            public override void VisitPrefixUnaryExpression(PrefixUnaryExpressionSyntax node)
+            {
+                base.VisitPrefixUnaryExpression(node);
+
+                CheckForFieldWrite(node.Operand);
+            }
+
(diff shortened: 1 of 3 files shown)


lgolding:
:+1:

src/Microsoft.DotNet.CodeFormatting/Rules/MarkReadonlyFieldsRule.cs
```diff
@@ -249,6 +249,20 @@ public override void VisitAssignmentExpression(AssignmentExpressionSyntax node)
...
+                base.VisitPostfixUnaryExpression(node);
+
+                CheckForFieldWrite(node.Operand);
+            }
+
+            public override void VisitPrefixUnaryExpression(PrefixUnaryExpressionSyntax node)
+            {
+                base.VisitPrefixUnaryExpression(node);
+
+                CheckForFieldWrite(node.Operand);
```
NF-Floyd: I think this should check for Increment operators, because the other unary operators like ! (negation) or ~ (bitwise complement) would be ok. I would do this like so:
```C#
            public override void VisitPrefixUnaryExpression(PrefixUnaryExpressionSyntax node)
            {
                base.VisitPrefixUnaryExpression(node);

                switch (node.OperatorToken.Kind())
                {
                    case SyntaxKind.MinusMinusToken:
                    case SyntaxKind.PlusPlusToken:
                        {
                            CheckForFieldWrite(node.Operand);
                            break;
                        }
                }
            }
```

michaelcfanning: Yes, nice catch.

michaelcfanning:
@NF-Floyd updated rule and added tests ensuring read-only still applied for non-mutating unary operators, +, -, !, ~, and cast.

michaelcfanning:
@genlu @lgolding please merge this

lgolding:
It has merge conflicts that need to be resolved before it can be merged. Did you want one of us to do that part?

Issue #185, opened 2015-10-04
explicit var rule resolves to 'object' where ImmutableArray<int>.Builder expected


michaelcfanning:
Some generics methods causing issues for explicit var check:
var result = ImmutableArray.CreateBuilder<int>();
is converted to 
            object result = ImmutableArray.CreateBuilder<int>();
where
            ImmutableArray<int>.Builder result = ImmutableArray.CreateBuilder<int>();
is expected.

PR #204, opened 2016-01-16
Fix bug in ProvideExplicitVariableTypeAnalyzer

genlu:
That is caused by not checking if the code has syntax/binding errors. This might fixes https://github.com/dotnet/codeformatter/issues/185 (I could only repro the bug by introducing errors, which might not be the original cause of this issue).

@michaelcfanning @srivatsn @lgolding @mavasani

### DIFF
--- a/src/Microsoft.DotNet.CodeFormatter.Analyzers/ProvideExplicitVariableTypeAnalyzer.cs
+++ b/src/Microsoft.DotNet.CodeFormatter.Analyzers/ProvideExplicitVariableTypeAnalyzer.cs
@@ -12,6 +12,8 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Options;
 
+using Microsoft.CodeAnalysis.CSharp.Symbols;
+
 namespace Microsoft.DotNet.CodeFormatter.Analyzers
 {
     [Export(typeof(DiagnosticAnalyzer))]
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.CodeFormatter.Analyzers/ProvideExplicitVariableTypeAnalyzer.cs
```diff
@@ -53,12 +53,15 @@ public override void Initialize(AnalysisContext context)
...
+                var model = syntaxContext.SemanticModel;
+                var token = syntaxContext.CancellationToken;
                 // C# syntax doesn't allow implicitly typed variables to have multiple declarators,
                 // but we need to handle error situations (incomplete or extra declarators).   
                 if (node.Type != null &&
                     node.Type.IsVar &&
                     !IsTypeObvious(node) &&
-                    !IsAnonymousType(node.Variables.FirstOrDefault(), syntaxContext.SemanticModel, syntaxContext.CancellationToken))
+                    !IsAnonymousType(node.Variables.FirstOrDefault(), model, token) &&
+                    !HasErrors(node, model, token))
```
srivatsn: So if there's any error anywhere in the program, we wouldn't apply this rule? That sounds too strict. I'd suggest simply checking if node.Type.IsErrorType() and bailing in that case.

Also computing all the diagnostics is going to be expensive as well.

genlu: Thanks, fixed now. By the way, the branch is still referencing Roslyn 1.0, I guess it can use an update ;)

michaelcfanning:
For sure this is #185. :) btw - has anyone discussed the core issue here, which appears to be that the 'remove unused namespaces' fixer is clobbering using directives that are, in fact, required to compile the code? 

... if not, then even after this change, code formatter will leave these repros in a non-buildable condition.

genlu:
@michaelcfanning So you are saying the "Remove unused using" fixer mistakenly removes (useful) usings which introduces the error that causes this issue?

michaelcfanning:
Yes, I think that's what's happening.

genlu:
@michaelcfanning I see. Could you please create a bug for that?

genlu:
I'd like to merge this if there's no objection. Guys?

srivatsn:
:+1:

PR #205, opened 2016-01-22
Fix ProvideExplicitVariableTypeAnalyzer

genlu:
Just noticed I didn't commit all the changes in my last PR (#204). Apologize for the bug introduced.

@srivatsn @michaelcfanning @lgolding


genlu:
Merging w/o review because the analyzer is apparently broken. Will address any comment after.

PR #207, opened 2016-02-09
Introduce local\global classification for rules based on analyzers

srivatsn:
Introduce the concept of local and global rules for analyzers such that all local fixes can be applied together whereas global rules will have to be applied one by one. 

This is necessary to fix an issue where the remove unused using fixer was removing an using that only became necessary by the "Make type explicit" fixer and they cannot be applied together.

This PR also implements the rule about placing usings outside the namespace as analyzers\fixers. The original rule in the Rules directory tries to simply move the using syntactically but that is incorrect. If there's using A inside namespace B then when moved out it might have to using B.A and so we need semantics. Fixing that in the analyzer and adding tests.

### DIFF
--- a/src/CodeFormatter/App.config
+++ b/src/CodeFormatter/App.config
@@ -37,6 +37,34 @@
         <assemblyIdentity name="Microsoft.CodeAnalysis.CSharp.Workspaces" publicKeyToken="31bf3856ad364e35" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-0.7.0.0" newVersion="0.7.0.0" />
       </dependentAssembly>
+      <dependentAssembly>  
+       <assemblyIdentity name="Microsoft.Build.Engine" publicKeyToken="b03f5f7f11d50a3a" culture="neutral"/>  
+       <bindingRedirect oldVersion="0.0.0.0-99.9.9.9" newVersion="14.0.0.0"/>  
+     </dependentAssembly>  
+     <dependentAssembly>  
+       <assemblyIdentity name="Microsoft.Build" publicKeyToken="b03f5f7f11d50a3a" culture="neutral"/>  
+       <bindingRedirect oldVersion="0.0.0.0-99.9.9.9" newVersion="14.0.0.0"/>  
+     </dependentAssembly>  
+     <dependentAssembly>  
+       <assemblyIdentity name="Microsoft.Build.Conversion.Core" publicKeyToken="b03f5f7f11d50a3a" culture="neutral"/>  
+       <bindingRedirect oldVersion="0.0.0.0-99.9.9.9" newVersion="14.0.0.0"/>  
+     </dependentAssembly>  
+     <dependentAssembly>  
+       <assemblyIdentity name="Microsoft.Build.Tasks.Core" publicKeyToken="b03f5f7f11d50a3a" culture="neutral"/>  
+       <bindingRedirect oldVersion="0.0.0.0-99.9.9.9" newVersion="14.0.0.0"/>  
+     </dependentAssembly>  
+     <dependentAssembly>  
+       <assemblyIdentity name="Microsoft.Build.Utilities.Core" publicKeyToken="b03f5f7f11d50a3a" culture="neutral"/>  
+       <bindingRedirect oldVersion="0.0.0.0-99.9.9.9" newVersion="14.0.0.0"/>  
+     </dependentAssembly>  
+     <dependentAssembly>  
+       <assemblyIdentity name="Microsoft.CompactFramework.Build.Tasks" publicKeyToken="b03f5f7f11d50a3a" culture="neutral"/>  
+       <bindingRedirect oldVersion="0.0.0.0-99.9.9.9" newVersion="9.0.0.0"/>  
+     </dependentAssembly>  
+     <dependentAssembly>  
+       <assemblyIdentity name="Microsoft.Build.Framework" publicKeyToken="b03f5f7f11d50a3a" culture="neutral"/>  
+       <bindingRedirect oldVersion="0.0.0.0-99.9.9.9" newVersion="14.0.0.0"/>  
+     </dependentAssembly>  
     </assemblyBinding>
   </runtime>
 </configuration>
--- a/src/Microsoft.DotNet.CodeFormatter.Analyzers/PlaceImportsOutsideNamespaceFixer.cs
+++ b/src/Microsoft.DotNet.CodeFormatter.Analyzers/PlaceImportsOutsideNamespaceFixer.cs
@@ -37,7 +37,7 @@ namespace Microsoft.DotNet.CodeFormatter.Analyzers
             foreach (Location location in diagnostic.AdditionalLocations)
             {
                 SyntaxNode usingDirectiveNode = root.FindNode(location.SourceSpan);
-                Debug.Assert(usingDirectiveNode != null);
+                Debug.Assert(usingDirectiveNode != null && usingDirectiveNode.IsKind(SyntaxKind.UsingDirective));
                 usingDirectiveNodes.Add(usingDirectiveNode);
             }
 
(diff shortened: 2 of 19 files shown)


srivatsn:
@lgolding @michaelcfanning @genlu @mavasani

srivatsn:
@dotnet/roslyn-analysis

src/CodeFormatter/App.config
```diff
@@ -37,6 +37,34 @@
         <assemblyIdentity name="Microsoft.CodeAnalysis.CSharp.Workspaces" publicKeyToken="31bf3856ad364e35" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-0.7.0.0" newVersion="0.7.0.0" />
       </dependentAssembly>
+      <dependentAssembly>
```
srivatsn: These binding redirects are needed to analyze some projects which have build tasks built against MSBuild 4.0. Since we will run with MSBuild 14.0, without these redirects those tasks would fail and we'll end up with a broken workspace otherwise.

src/Microsoft.DotNet.CodeFormatter.Analyzers/PlaceImportsOutsideNamespaceFixer.cs
```diff
@@ -0,0 +1,97 @@
...
+            var diagnostic = context.Diagnostics.First();
+            var usingDirectiveNodes = new List<SyntaxNode>();
+
+            // We recapitulate the primary diagnostic location in the 
+            // Diagnostic.AdditionalLocations property on raising
+            // the diagnostic, so this member has complete location details.
+            foreach (Location location in diagnostic.AdditionalLocations)
+            {
+                SyntaxNode usingDirectiveNode = root.FindNode(location.SourceSpan);
+                Debug.Assert(usingDirectiveNode != null);
```
lgolding: Do you also want to assert that it really is a `UsingDirectiveSyntax`?

src/Microsoft.DotNet.CodeFormatting/RuleType.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// <summary>
+        /// This type of rule is a purely syntactic rule and doesn't need any semantic information to analyze or fix the issue.
+        /// Such rules can be correct in a broken compilation as well.
+        /// </summary>
+        public const string Syntactic = nameof(Syntactic);
+
+        /// <summary>
+        /// This type of rule needs to look at semantic information to either diagnose or fix the issue. The semantic information
+        /// is localized the site of the issue and any fix will only semantically affect the site of the issue or the enclosing code block 
+        /// for eg: an edit inside a method body.
```
lgolding: "for eg" +. either "for example" or "e.g."

src/Microsoft.DotNet.CodeFormatting/FormattingEngineImplementation.cs
```diff
@@ -179,10 +180,18 @@ public ImmutableArray<DiagnosticDescriptor> AllSupportedDiagnostics
...
         public async Task FormatSolutionWithAnalyzersAsync(Solution solution, CancellationToken cancellationToken)
         {
-            foreach (var project in solution.Projects)
+            var watch = new Stopwatch();
+            watch.Start();
+
+            var workspace = solution.Workspace;
+            foreach (var projectId in solution.ProjectIds)
             {
+                var project = workspace.CurrentSolution.GetProject(projectId);
```
lgolding: What's the reason for this change? (Maybe I'll find out as I read further...)

src/Microsoft.DotNet.CodeFormatting/FormattingEngineImplementation.cs
```diff
@@ -196,7 +205,43 @@ public ImmutableArray<DiagnosticDescriptor> AllSupportedDiagnostics
...
+            await FormatProjectWithLocalAnalyzersAsync(workspace, project.Id, cancellationToken);
+            await FormatProjectWithGlobalAnalyzersAsync(workspace, project.Id, cancellationToken);
+
+            watch.Stop();
+            FormatLogger.WriteLine("Total time for formatting {0} - {1}", project.Name, watch.Elapsed);
+        }
+
+        private async Task FormatProjectWithSyntaxAnalyzersAsync(Workspace workspace, ProjectId projectId, CancellationToken cancellationToken)
+        {
+            var analyzers = _analyzers.Where(a => a.SupportedDiagnostics.All(d => d.CustomTags.Contains(RuleType.Syntactic)));
```
lgolding: Yup, this is what I was worried about. This means that unless an analyzer explicitly declares its scope (`RuleType`), CodeFormatter won't run it at all. We had originally hoped to be able to pass arbitrary analyzer DLLs on the codeformatter command line so end users could say "please format my project with these style analyzers", but with this design, only analyzers that know enough to set these particular custom tags can run.

One option is to follow this up by implementing command line options like
```
--analyzer-assemblies foo.dll,bar.dll --rules SA1001,SA1002
```
or
```
--analyzer-assemblies foo.dll,bar.dll --all-rules
```
... which would allow the user to say "I know what I'm doing, just run these".

src/Microsoft.DotNet.CodeFormatting/FormattingEngineImplementation.cs
```diff
@@ -316,7 +357,31 @@ private Solution RemoveTablePreprocessorSymbol(Solution newSolution, Solution ol
...
+                {
+
+                        Console.WriteLine("Error count " + compilerDiagnostics.Count());
+
+                        Console.WriteLine("Metadata References for {0}", project.Name);
+                        foreach (var mr in project.MetadataReferences)
+                        {
+                            Console.WriteLine(mr.Display);
+                        }
+
```
lgolding: Maybe want some separator text or white space between the metadata and the diagnostics?

src/Microsoft.DotNet.CodeFormatting/FormattingEngineImplementation.cs
lgolding: (Actually, no I never did figure out why this change was needed.)

lgolding:
:+1: with some comments.

src/Microsoft.DotNet.CodeFormatting/FormattingEngineImplementation.cs
srivatsn: The Format method applies a codefix which applies changes to the workspace. That means that the currentsolution of the workspace would get updated. So unless we get the project from the current solution, we'd be formatting a Project from a previous solution and Workspace.ApplyChanges would reject the changes to that Project (because it's not to the latest snapshot). The net effect is that without this change only the first project would get formatted.

src/Microsoft.DotNet.CodeFormatting/FormattingEngineImplementation.cs
srivatsn: Yes I had forgotten about this aspect. When we add the switch we could choose to be either conservative (if an analyzer doesn't have one of these tags then just treat is as a GlobalSemantic analyzer and apply them one by one) or be very lax (treat them as LocalSemantic analyzers and apply them together and they could break the code).

src/Microsoft.DotNet.CodeFormatting/FormattingEngineImplementation.cs
srivatsn: Added

src/Microsoft.DotNet.CodeFormatter.Analyzers/ProvideExplicitVariableTypeAnalyzer.cs
```diff
@@ -30,14 +30,14 @@ public class ProvideExplicitVariableTypeAnalyzer : DiagnosticAnalyzer
...
+                                                                            customTags: new[] { VariableDeclarationCustomTag, RuleType.LocalSemantic });
         private static readonly DiagnosticDescriptor s_ruleForEachStatement = new DiagnosticDescriptor(DiagnosticId,
                                                                             ResourceHelper.MakeLocalizableString(nameof(Resources.ExplicitVariableTypeAnalyzer_Title)),
                                                                             ResourceHelper.MakeLocalizableString(nameof(Resources.ExplicitVariableTypeAnalyzer_MessageFormat)),
                                                                             "Style",
                                                                             DiagnosticSeverity.Warning,
                                                                             true,
-                                                                            customTags: ForEachStatementCustomTag);
+                                                                            customTags: new[] { ForEachStatementCustomTag, RuleType.LocalSemantic });
```
genlu: On a completely unrelated note: is there a plan to upgrade Roslyn reference? I'd like to fix this multiple descriptor issue here by using the new property bag in `Diagnostic`.

PR #208, opened 2016-02-10
Fixing a bug in the Provide Explicit type rule where it was trying to…

srivatsn:
… make a generic type with anonymous type parameters explicit.

There was also a case where if a foreach statement had an error type, we weren't bailing out correctly.

@lgolding @genlu @mavasani

### DIFF
--- a/src/Microsoft.DotNet.CodeFormatter.Analyzers.Tests/ProvideExplicitVariableTypeAnalyzerTests.cs
+++ b/src/Microsoft.DotNet.CodeFormatter.Analyzers.Tests/ProvideExplicitVariableTypeAnalyzerTests.cs
@@ -300,6 +300,47 @@ class C1
         }
 
         [Fact]
+        public void TestDeclarationWithAnonymousTypeArguments()
+        {
+            const string input = @"
+using System.Linq;
+
+class C1
+{
+    void M()
+    {
+        var names = from i in new [] { 1, 2, 3 }
+                    select new { Number = i, Age = i };
+
+        foreach (var name in names)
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.CodeFormatter.Analyzers.Tests/ProvideExplicitVariableTypeAnalyzerTests.cs
```diff
@@ -300,6 +300,47 @@ void M()
...
+        {
+            const string input = @"
+using System.Linq;
+
+class C1
+{
+    void M()
+    {
+        var names = from i in new [] { 1, 2, 3 }
+                    select new { Number = i, Age = i };
```
lgolding: I see. The type of the variable `names` is `IEnumerable<some_anonymous_type>` -- so `names` is not itself of anonymous type; rather, it is a generic type whose type parameter is anonymous. It took me a minute to understand  how this example tested your new code. Perhaps a comment:

```
// The variable names if of type IEnumerable<some_anonymous_type>. It is not itself of anonymous type,
// but it has a type parameter of anonymous type.

lgolding:
:+1:

src/Microsoft.DotNet.CodeFormatter.Analyzers.Tests/ProvideExplicitVariableTypeAnalyzerTests.cs
srivatsn: Done

PR #210, opened 2016-02-11
The rule that makes types explicit should add a using when it needs to qualify type.

srivatsn:
If there's code that's being converted from var to an explicit type and there's no using for that type in the file, the fix should add the using instead of generating a fully qualified name.

@lgolding @genlu @mavasani

### DIFF
--- a/src/Microsoft.DotNet.CodeFormatter.Analyzers/ProvideExplicitVariableTypeFixer.cs
+++ b/src/Microsoft.DotNet.CodeFormatter.Analyzers/ProvideExplicitVariableTypeFixer.cs
@@ -80,8 +80,11 @@ namespace Microsoft.DotNet.CodeFormatter.Analyzers
         private async Task<Document> ReplaceVarWithExplicitType(Document document, SyntaxNode varNode, ITypeSymbol explicitTypeSymbol, CancellationToken cancellationToken)
         {
             DocumentEditor documentEditor = await DocumentEditor.CreateAsync(document, cancellationToken);
+
+            // Create an annotation and tag the type node with it so that we can find it again in a new tree.
+            var explicitTypeAnnotation = new SyntaxAnnotation();
             SyntaxNode explicitTypeNode = documentEditor.Generator.TypeExpression(explicitTypeSymbol)
-                                          .WithAdditionalAnnotations(Simplifier.Annotation)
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.CodeFormatter.Analyzers/ProvideExplicitVariableTypeFixer.cs
```diff
@@ -84,7 +84,15 @@ public override FixAllProvider GetFixAllProvider()
...
                                           .WithTriviaFrom(varNode);
             documentEditor.ReplaceNode(varNode, explicitTypeNode);
-            return documentEditor.GetChangedDocument();
+            var newDocument = documentEditor.GetChangedDocument();
+
+            // We don't want the explicit type to be fully qualified. So add an using for this node and the simplifier will
+            // take care of removing it if it isn't necessary.
+            // The second parmaeter to AddImportsAsync is an annotation that is used to locate the span
+            // for which an using should be added. Since we added the simplify annotation on explicitTypeNode, 
+            // we can just use that annotation to locate the node again in newDocument.
```
lgolding: If I understand this right, the following call says: to the specified document, add any imports that are required by the node with the "simplifier" annotation. And then later on, the fact that this node _has_ the simplifier annotation will cause the fully qualified type name to be replaced with a simple type name.

Can this go wrong in any way? Is it possible that some node could have a simplifier annotation without us having put it there? If so, we could invent our own "provide explicit variable type" annotation, and add that to the node _in addition to_ the simplifier annotation. Then here, we could pass this annotation to `AddImportAsync` (instead of passing the simplifier annotation), secure in the knowledge that the only nodes with this annotation are the ones we explicitly marked ourselves?

lgolding:
:+1: (assuming that the answer to my comment is "no, don't worry, that can't happen")

genlu:
:+1: , other than @lgolding 's concern.

src/Microsoft.DotNet.CodeFormatter.Analyzers/ProvideExplicitVariableTypeFixer.cs
srivatsn: In the current implementation, the simplifier is called immediately (during PostProcess) on a codefix and the given document shouldn't have any simplifier annotations. Having said that I was being lazy here and depending on an implementation detail. Fixed it to create a separate annotation instead.

PR #211, opened 2016-02-11
Revert "The rule that makes types explicit should add a using when it needs to qualify type."

srivatsn:
Reverts dotnet/codeformatter#210

It turns out that ImportAdder API in Roslyn has a bug where it doesn't check if adding the using causes any ambiguity errors. This causes this rule to break code. So reverting this change until the Roslyn bug can be addressed.


srivatsn:
FYI @lgolding @genlu 

If you are curious

lgolding:
Fascinating :-)

genlu:
Wow, this is hard :)

PR #212, opened 2016-02-17
Add a rule to collapse multiple new lines to a single newline.

paulvanbrenk:
Feedback?


paulvanbrenk:
Going to do this differently

PR #217, opened 2016-03-21
Allow codeformatter to use analyzers from external DLLs

danquirk:
No corresponding issue for this as far as I know but I can create one if you'd like. I've talked to @srivatsn about this a bit in generalities.

The basic idea is to pass codeformatter a DLL which contains some analyzers and it will run those analyzer(s) against the given target(s). The targets can be project path or a file listing project path and the analyzers can be a path to the DLL or a file listing paths to multiple DLLs. A new ```analyze``` command is added to do only analysis of the target without applying any corresponding code fixes from the analyzers.

I had to upgrade to newer versions of Roslyn and System.Reflection.Metadata to support the latest analyzers from Roslyn and 3rd parties. I had tried to add new analyzers just using MEF rather than IAnalyzerFileReference (see [this branch](https://github.com/danquirk/codeformatter/tree/MEFAnalyzersV2)) which seemed like it should be cleaner but didn't seem to work for certain analyzers. There's some unfortunate reflection into Roslyn for internal types that may not be the fastest/best solution long term.

TODO:
   - [ ] Add some more tests
   - [ ] Log analyzer execution time in SARIF format
   - [ ] Test VB


src/CodeFormatter/Program.cs
```diff
@@ -117,11 +131,42 @@ private static int RunFormatCommand(FormatOptions options)
...
-        private static async Task<int> RunFormatAsync(FormatOptions options, CancellationToken cancellationToken)
+        private static IAnalyzerAssemblyLoader _analyzerAssemblyLoader;
+        private static IAnalyzerAssemblyLoader AnalyzerAssemblyLoader {
+            get
+            {
+                if(_analyzerAssemblyLoader == null)
+                {
+                    // use Roslyn's existing simple loader, there're no special requirements for our usage
+                    var loaderAssembly = Assembly.Load((typeof(CommandLineProject)).Assembly.FullName);
+                    _analyzerAssemblyLoader = (IAnalyzerAssemblyLoader)Activator.CreateInstance(loaderAssembly.GetType("Microsoft.CodeAnalysis.SimpleAnalyzerAssemblyLoader"));
```
srivatsn: Do you really need Roslyn's assembly loader? That does a bunch of caching and hooks assembly resolve to be able to find dependencies from any of the analyzer folders. You could impose a simpler model here and have assembly loader simply do a LoadFrom(..) and just require all dependencies to be in the same folder. That would avoid reflection here as well.

danquirk: Good question, I'll investigate. I originally started to write my own trivial one and then saw the Roslyn version and figured 'SimpleAnalyzerAssemblyLoader' sounded like just what I needed :)

danquirk: Done.

src/Microsoft.DotNet.CodeFormatting.Tests/FormattingEngineCreationTests.cs
```diff
@@ -0,0 +1,66 @@
...
+    public class FormattingEngineCreationTests
+    {
+
+        private static Assembly[] DefaultCompositionAssemblies =
+                                        new Assembly[] {
+                                            typeof(FormattingEngine).Assembly,
+                                            typeof(OptimizeNamespaceImportsAnalyzer).Assembly
+                                        };
+        // TODO: fix hardcoded path
+        private static string TestDLLDir = @"E:\src\codeformatter\src\Microsoft.DotNet.CodeFormatting.Tests\TestAnalyzers\";
```
mavasani: Did you miss fixing this TODO?

danquirk: Yep. Should've put 'Ctrl-F for TODO in my list of TODOs' ;)

Been sidetracked with other work the past week or so but will be revisiting this and cleaning it up with any other CR comments in the next day or two, thanks for looking.

src/CodeFormatter/Program.cs
```diff
@@ -132,7 +163,33 @@ private static int RunFormatCommand(FormatOptions options)
...
+            {
+                foreach (var analyzerPath in options.AnalyzerListText)
+                {
+                    if (File.Exists(analyzerPath))
+                    {
+                        var newAnalyzers = LoadAnalyzersFromAssembly(analyzerPath, true);
+                        engine.AddAnalyzers(newAnalyzers);
+                    }
+                    else if (Directory.Exists(analyzerPath))
+                    {
```
basoundr: Add a test this for as well?

basoundr:
LGTM pending the testcase

basoundr:
@danquirk Do you plan on addressing the TODO in this PR or a follow-up PR? If you are planning on doing a follow-up PR then we can merge this change in

danquirk:
@basoundr unless you have an immediate need for it in its current form I think I'd prefer to just do the fixes in this PR and then have it all merged at once in a better state. I'll try to get to these other review comments shortly.

basoundr:
Our team is having a FxCop day(All our team members work on just the analyzers-related things on this day) this Thursday. We were hoping that, if you could have this change merged in and have CodeFormatter run on repos like CoreFx & others, which are amazing testbeds for analyzers, then we might hit some bugs in the Analyzers and have those bugs addressed during our FxCop day. Is this something that would fit into your schedule?

danquirk:
Yeah, that seems doable.

basoundr:
Awesome! I will merge this PR in and I will create an issue with the TODO mentioned in the PR

basoundr:
@danquirk I have created the issue #225 for you to track the TODOs. I am not able to assign the issue to you, so please assign it yourself.

Issue #151, opened 2015-07-27
/nocopyright should be the default


hickford:
// Copyright (c) Microsoft. All rights reserved.
    // Licensed under the MIT license. See LICENSE file in the project root for full license information.

This header is inappropriate for most projects (outside Microsoft, other licenses). To be useful to more people `codeformatter.exe` shouldn't make these assumptions out the box. 

Simplest solution: `/nocopyright` should be the default.

hickford:
Workaround: create an empty file `empty.txt` then run `codeformatter.exe /copyright:empty.txt`

PR #220, opened 2016-03-30
Small cleanup

davkean:
Some small cleanup which will make my fix for #151 be less churn.


davkean:
tag @jaredpar @terrajobst

terrajobst:
LGTM

@Priya91

PR #221, opened 2016-03-31
Turned copyright header off by default

davkean:
Fixes: #151.

Before my change there were two options, specifying `/nocopyright`, or specifying `/copyright:foo.txt`. If we switched the header off by default, there was no way to opt into the default, therefore, I added enable/disable options to the copyright switch similar to /rule:

To turn on default copyright, you do:

```
codeformatter test.proj /copyright
codeformatter test.proj /copyright+
```

To turn off default copyright, you do:

```
codeformatter test.proj
codeformatter test.proj /copyright-
codeformatter test.proj /nocopyright (for backwards compat - not doc'd)
```

And like before you can specify a file:
```
codeformatter test.proj /copyright:copyright.txt
```

### DIFF
--- a/src/CodeFormatter/CommandLineParser.cs
+++ b/src/CodeFormatter/CommandLineParser.cs
@@ -194,13 +194,13 @@ namespace CodeFormatter
                         return CommandLineParseResult.CreateError(error);
                     }
                 }
-                else if (arg.StartsWith(LanguageSwitch, comparison))
-                {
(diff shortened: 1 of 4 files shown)


davkean:
tag @jaredpar @terrajobst 

Make note I pulled in https://github.com/dotnet/codeformatter/pull/220, so first two commits will disappear.

davkean:
ping

terrajobst:
That makes sense. LGTM.

Issue #225, opened 2016-04-12
Address PR Comments for #217


basoundr:
The PR #217 added the ability to run CodeFormatter with Analyzers. This issue has been created to finish the TODO mentioned in the PR.

TODO:
   - [ ] Add some more tests
   - [ ] Log analyzer execution time in SARIF format
   - [ ] Test VB

danquirk:
I'm working on this but can't formally assign it.

PR #227, opened 2016-04-14
Custom analyzer fixes for PR 217

danquirk:
For https://github.com/dotnet/codeformatter/issues/225

- First commit fixes a bug I discovered in the command line options that were added. This was broken but will now work:
```codeformatter.exe analyze <targetProjFile> --use-analyzers --analyzers <pathToAnalyzerDLL>```
and this continues to work:
```codeformatter.exe analyze <targetProjFile> --use-analyzers --analyzers <pathToTxtFileWithAnalyzerDLLs>```

- Tests have been rewritten to correctly test the non-MEF based analyzer loading that was checked in. However only 1 of 5 passes for reasons I'm having trouble debugging (tests for assemblies besides the 1.2.0-beta1 version claim to have no analyzers in it). The scenarios they are meant to test work correctly on the command line though (manually running against those test analyzer assemblies produces analyzer diagnostics). 

- Log analyzer execution time to the console if desired. The Roslyn ErrorLogger doesn't make it easy to insert that information in SARIF format at the moment. We can add it later as necessary.

Tagging @basoundr and @srivatsn. You'll definitely want the fix in the first bullet for playing with this tomorrow.


srivatsn:
👍

Issue #174, opened 2015-09-04
keep tabs instead of spaces


BertTank:
I know this is your default and written down in the documentation: ```We use four spaces of indentation (no tabs)``` but we use tabs by default in my company. However the rest of your rules will perfectly fit ours.

Is it possible to exclude this feature to keep tabs (or even convert spaces to tabs)?

genlu:
@BertTank I believe you can disable the tab formatting by adding `/rule-:FormatDocument` at commandline.

BertTank:
That's right, but I want to have the formatting of that rule - just not with spaces but with tabs.

genlu:
@BertTank Yeah, @michaelcfanning is working on adding solution/project specific options to formatting rules. This space-tab conversion option will definitely be considered.

PR #229, opened 2016-04-14
Add ability to use tabs

michalhosala:
Added functionality which addresses issue #174, i.e. ability to use tabs instead of spaces during formatting. New command line switch /usetabs was introduced for this purpose.

### DIFF
--- a/src/CodeFormatter/CommandLineParser.cs
+++ b/src/CodeFormatter/CommandLineParser.cs
@@ -28,7 +28,8 @@ namespace CodeFormatter
             ImmutableArray<string>.Empty,
             null,
             allowTables: false,
-            verbose: false);
+            verbose: false,
+            useTabs: false);
 
         public readonly Operation Operation;
         public readonly ImmutableArray<string[]> PreprocessorConfigurations;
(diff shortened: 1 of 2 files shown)


anderejd:
Is this project abandoned?

tiesmaster:
@anderejd See issue #267, specifically, [davkean's reply](https://github.com/dotnet/codeformatter/issues/267#issuecomment-331320957). I think this is still being used internally in MSFT, but no support anymore :(

I know that Roslyn has it's own [`$ dotnet format`](https://github.com/dotnet/roslyn/tree/master/src/Tools/dotnet-format) tool ([.NET Core global tool](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools)), which is still very experimental, if I'm not mistaken. I hope this becomes the "new `codeformatter`" out there.

anderejd:
@tiesmaster Excellent! Thank you so much, I had no idea of the existence of `dotnet-format`. With an `.editorconfig` like this it will tabify C#:
```
[*.cs]
indent_style = tab
```

PR #237, opened 2016-06-16
Upgrade Codeformatter to Roslyn 1.3-beta1

danquirk:
Need to upgrade so that we can use the latest ErrorLogger from Roslyn which meets the 1.0 SARIF spec requirements.

### DIFF
--- a/src/Microsoft.DotNet.CodeFormatting/FormattingEngineImplementation.cs
+++ b/src/Microsoft.DotNet.CodeFormatting/FormattingEngineImplementation.cs
@@ -107,12 +107,17 @@ namespace Microsoft.DotNet.CodeFormatting
 
             _outputLogger = Activator.CreateInstance(
                 _loggerAssembly.GetType("Microsoft.CodeAnalysis.ErrorLogger"),
-                new object[] { new FileStream(filePath, FileMode.Append, FileAccess.Write), "CodeFormatter", "0.1", _loggerAssembly.GetName().Version });
+                new object[] {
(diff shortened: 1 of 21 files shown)


src/Microsoft.DotNet.CodeFormatting/FormattingEngineImplementation.cs
```diff
@@ -107,12 +107,17 @@ public void LogDiagnostics(string filePath, ImmutableArray<Diagnostic> diagnosti
 
             _outputLogger = Activator.CreateInstance(
                 _loggerAssembly.GetType("Microsoft.CodeAnalysis.ErrorLogger"),
-                new object[] { new FileStream(filePath, FileMode.Append, FileAccess.Write), "CodeFormatter", "0.1", _loggerAssembly.GetName().Version });
-
-            var logDiagMethodInfo = _outputLogger.GetType().GetMethod("LogDiagnostic", BindingFlags.NonPublic | BindingFlags.Instance);
+                new object[] {
+                    new FileStream(filePath, FileMode.Append, FileAccess.Write),
+                    "CodeFormatter",
+                    "0.1",
```
lgolding: This string ends up being persisted to the SARIF log file, which requires it to have the format `"[0-9]+(\\.[0-9]+){3}"` (it's intended to be a .NET FileVersion attribute value). So you could get it from `Assembly.GetEntryAssembly`.

danquirk: Ah, ok, will do.

Issue #106, opened 2015-04-10
Add ability to run on OS X / Xamarin Studio


mteper:
Would be nice if this could be used outside Windows.

AndyGerlicher:
In the xplat branch, we've been using monodevelop / Xamarin Studio. Have you tried that branch? And if so what issues are you seeing?

mteper:
Nope, haven't tried yet, wasn't aware of it. Thanks for the pointer!

AndyGerlicher:
I'll close this then. Reopen if you have any issues.

AndyGerlicher:
Uhg, I was working in the msbuild repo and clicked on this issue link in my email... There's definitely not an xplat branch or Linux support here, so I'll reopen. Sorry about that!

ghuntley:
Found out about this project early this morning and was like all

![](http://24.media.tumblr.com/tumblr_lzctt8R7gE1qdg9v3o9_400.gif)

that is until:

    › mono CodeFormatter.exe ../../src/SuperSecret.sln
    ERROR: Type loading error detected. In order to run this tool you need either 
    Visual Studio 2015 or Microsoft Build Tools 2015 tools installed.
    - Could not load file or assembly or one of its dependencies.

A while back I took on ownership of a fork for ```narrange``` and wrote a corresponding plugin for  monodevelop. I would be keen in working with someone (@mteper)? to make this a first class citizen.

@AndyGerlicher when will the ```xplat``` branch be published?

![](http://i.imgur.com/P88UGIT.gif)

jaredpar:
I think cross plat is a goal we should definitely shoot for on this tool.  The code in the tool itself should be fine for cross plat.  It may need a little work but nothing series.  The bigger issue is our dependencies.  

The Roslyn code uses the MSBuild APIs to create a Solution object from the provided solution and project files.  The MSBuild DLLs, and the Roslyn DLLs which depend on it, are not designed to be cross platform.  I imagine it will take some level of work to get this running cross platform.

The good news though is MSBuild is moving to cross plat support so this should be a very doable proposition for us.  It will probably take a bit of effort though.  

Another option is to have a cross plat version of the tool which only supports the rsp command line arguments.  This doesn't need cross plat support to read in and hence should in theory just work.  

Either way, something to explore.  

Note: in the future could we have less GIFs in the thread?  Having a toddler swaying back and forth on my screen is simply distracting when trying to type up responses to issues :smile:

mteper:
[...]

Issue #199, opened 2015-11-27
Support DNX


jskeet:
This is a sort of bigger feature request version of issue #172.

In the brave new world, DNX-based projects and solutions are going to be more and more common. I haven't yet figured out a way of telling `codeformatter` to format all the code within a DNX solution or DNX project, as you can't specify individual source files (entirely reasonably, given that it's doing more than formatting).

jaredpar:
This is definitely in the works now. The intent is to add support for project.json based projects. A couple of other pieces need to fall into place before we get full parity with MSBuild based project. But this will be coming.

Sent using OWA for iPhone
________________________________

davidfowl:
The workspace implementation should be good enough (hopefully) https://github.com/dotnet/cli/tree/master/src/Microsoft.DotNet.ProjectModel.Workspaces. We need to start producing packages for these so we can consume them here.

stuartleeks:
Now that VS2017 is released, can we get a version of code formatter that will work on .NET Core based csproj projects (#256), and on a machine with only VS2017 installed, i.e. where .NET 4.5.2 isn't installed (#255)?

jaredpar:
@stuartleeks it's on my radar to do.  Still a lot of VS 2017 ship work going on right now.  Once that calms down though going to be taking a look at this.

stuartleeks:
I'll keep my fingers crossed that things calm down soon then ;-)

gilescope:
.net core formatter would be great...

PR #245, opened 2016-11-11
allow CodeFormatter to run on Mono

alexsorokoletov:
I would love to have a single source of truth for formatting C# code without any options to customize it. 

Turns out there is a CodeFormatter project (thank you!) but it doesn't run well on Mono. During MVPSummit Hackathon with **great help of @brettfo and guidance from @jaredpar ** we've been able to allow to run the CodeFormatter on mono with little changes.

How to run it on Mono then?
First, install the tools `sh init-tools.sh` (downloads and unzips MSBuild for Mono, one could possibly symlink to current install instead of downloading, idea taken from [Mono's MSBuild fork xplat-c9](https://github.com/mono/msbuild/blob/5c03c87120a7e8f03bbe40a386c738bbbe54ba4e/cibuild.sh)).

Restore NuGet packages and build the project: 
`nuget restore src/CodeFormattter.sln && msbuild src/CodeFormatter.sln` 

Then run it:
`mono bin/CodeFormatter/Debug/CodeFormatter.exe ../codeformatter-test/test.rsp`

Here's short video on overall process https://youtu.be/orPvC8y-Km4

Please let me know if there is a better way, I'm open for feedback.
The pull request is related to #106

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -235,3 +235,6 @@ Icon
 Network Trash Folder
 Temporary Items
 .apdisk
+
+# Local msbuild copy to run on Mono
+src/ExternalApis/
--- a/init-tools.sh
+++ b/init-tools.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+SRC="src"
+EXTERNAL_APIS_DIR="$SRC/ExternalApis"
+#see https://github.com/mono/msbuild/blob/xplat-c9/cibuild.sh
+MSBUILD_DOWNLOAD_URL="https://github.com/radical/msbuild/releases/download/v0.03/mono_msbuild_d25dd923839404bd64cc63f420e75acf96fc75c4.zip"
+MSBUILD_ZIP="$EXTERNAL_APIS_DIR/msbuild.zip"
+MSBUILD_EXE="$EXTERNAL_APIS_DIR/MSBuild/msbuild.exe"
+
+downloadMSBuildForMono()
+{
+    if [ ! -e "$MSBUILD_EXE" ]
+    then
+        mkdir -p $EXTERNAL_APIS_DIR
+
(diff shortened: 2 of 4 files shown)


perlun:
Any chance this could work with .NET Core nowadays? Or Mono only?

Petermarcu:
@jaredpar , do you know offhand what dependencies may be blocking this on Core?

alexsorokoletov:
@Petermarcu @perlun not Jared, but I believe it should work just fine on .NET Core. We would need to compile it against .NET Core and use MSBuild from .NET Core

jaredpar:
@Petermarcu it could move to .Net core with a little bit of work as @alexsorokoletov suggested.  I think there is a small amount of design work involved but probably 1-2 days of work.

perlun:
Sounds great! Could anyone create a GH issue about it, so we can track progress (and design ideas) there?

TimonVS:
What's the status on this issue? After using Prettier and gofmt it's quite hard to go back to a programming language that doesn't have a formatter 😅

alexsorokoletov:
Jared would be best person to outline the plan, but I could not agree more with Timon - after using formatter it's hard not to use it.

Basically, we have to migrate it to .net core and it will run anywhere where .net core runs.

alexsorokoletov:
Here are few similar issues already opened:
- https://github.com/dotnet/codeformatter/issues/199
- https://github.com/dotnet/codeformatter/issues/256
- https://github.com/dotnet/codeformatter/issues/106

I've quickly tried to migrate to .NET Core and there one thing that stops it:
https://github.com/dotnet/corefx/issues/18452#issuecomment-305009282

Formatting project uses package `Microsoft.CodeAnalysis.CSharp.Workspaces` and it references the `Microsoft.Composition` package not available on .NET Core (should be migrated to System.Composition) and there a some changes related to that (MEF classes and namespaces mostly).

PR #101, opened 2015-03-26
Add rule to remove unused usings.

david-mitchell:
This pull request adds a rule to remove unused using directives from the beginning of source files. It addresses issue #79.

Issue #79, opened 2015-02-24
Unused usings should be removed


jterry75:
If a file contains a using declaration that is not actually used it should be removed.

```C#
using System;
using System.Collections.Generic;

class Test
{
    public static void Main()
    {
        Console.WritLine("not using generics");
    }
}
```

david-mitchell:
I would like to take this one.

From my understanding, support for this would require a new `ILocalSemanticFormattingRule` that would run the diagnostics looking for `CS8019` warnings and removing the offending `using` directives. As in the `UsingLocationRule`, files with preprocessor directives should probably be ignored.

Daniel-Svensson:
It would be great if you could specify some namespaces which should always be kept such as System.Linq.

PR #278, opened 2018-10-13
Add rule to remove unused usings. #79

amanda-mitchell:
This is an update to #101, which was opened about three years ago and seems to have died of neglect. I've cleaned up the merge conflicts that crept in over that time period so that it could be resubmitted.


amanda-mitchell:
Closing since it seems pretty clear after five years that this project isn't terribly open to external contributions.

